PR #31, opened 2015-01-11
update autohint Xcode project files

schriftgestalt:
This was done on MacOSX Yosemite with Xcode 6 but tested with Xcode 5, too.
fixed issues:
- The name of the library subproject included the hole path.
- Framework search removed that caused problems with 10.10 SDK
- Use standard CONFIGURATION_BUILD_DIR, not the directory in the source folder
- Simplify the preprocessor macros a bit


FDK/Tools/Programs/autohint/build/osx/xcode4/autohintexe.xcodeproj/project.pbxproj
```diff
@@ -58,6 +43,7 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
+		56AE0BA71A633867000ECBEE /* libautohintlib.a */ = {isa = PBXFileReference; lastKnownFileType = archive.ar; name = libautohintlib.a; path = "../../../../../../../../../Library/Developer/Xcode/DerivedData/autohintexe-aqaujhwuqqtoouhccdzyzjbhhmcb/Build/Products/Debug/libautohintlib.a"; sourceTree = "<group>"; };
```
anthrotype: @readroberts the path to "libautohintlib.a" referenced here is only temporary and specific to @schriftgestalt Xcode setup at the time he edited the xcodeproj. 
Could that be the reason why I cannot successfully run "BuildAll.sh" fo autohint?
This is clang's error:

    ld: library not found for -lautohintlib  

The only way I can build autohint on OS X is by doing this:
- build `libautohintlib.a` using its own Xcode project `autohintlib.xcodeproj`;
- copy the resulting `autohint/autohintlib/exe/osx/debug/libautohintlib.a` to both `autohint/exe/osx/debug` and to `autohint/exe/osx/release`;
- finally, run autohint's BuildAll.sh script

schriftgestalt:
I’ll have a look a the build settings. 
Who and why would you call `BuildAll.sh` instead of using the targets in Xcode. I only use Xcode directly to build things (that explains why my changes broke the script build path.

anthrotype:
thanks Georg.
I find the BuildAll.sh scripts quite convenient, especially the one found in the root of the Programs folder, that can build all the akfdo tools at once from the command line.

PR #43, opened 2015-03-11
Update autohint.py

benkiel:
Typo found while verifying which version of the FDK I had installed. Though, I'd love to see the 2105 version of auto hint, perhaps by then it will be a single line: pass.


readroberts:
Thanks! My hope is also that by 2105, all displays will be the equivalent of at least 1200 dpi and hinting will be unnecessary,

kenlunde:
@readroberts I think that by 2105, 90 years from now, all displays may be 12,000 dpi or better and hinting will have gone the way of the dinosaurs. ;-)

(Sorry. I couldn't resist. ☺)

benkiel:
@readroberts @kenlunde It will be known as the "Great Hint Code Extinction of 2100"

Issue #48, opened 2015-04-17
Mark-to-base positioning rule error


jamesgk:
When running makeotf with a feature file containing a mark-to-base positioning rule, I always get the following error (or something like it):

```
makeotfexe [ERROR] <My-Font> MarkToBase or MarkToMark error 2: A previous statement has already assigned the current mark class to another anchor point on the same glyph 'A'. Skipping rule.  [features.fea 7]
```

A trivial example feature definition which causes this error:
```
languagesystem DFLT dflt; 
         
feature mark {
             
    lookup mark1 {
        markClass uni0300 <anchor 0 509> @mark_class;
        pos base A <anchor 280 678> mark @mark_class;                                          
    } mark1;
         
} mark;
```

This error does not occur with the pre-compiled FDK, just with the latest code from GitHub.

jamesgk:
@roozbehp @behdad

jamesgk:
This appears to be caused by a20059e5347d546dd68fccb2cb32b7327bd496b7, which breaks code in [GPOS.c](https://github.com/adobe-type-tools/afdko/blob/master/FDK/Tools/Programs/makeotf/makeotf_lib/source/hotconv/GPOS.c), as it assumes `-1 == 0xFFFFFFFFL`. The consensus in the discussion for issue #17 seems to be that 32-bit dependence is bad, so the FDK probably should not have code making this assumption (though reverting a20059e5347d546dd68fccb2cb32b7327bd496b7 does fix this error). Not sure at all that GPOS.c is the only problem area.

jenskutilek:
I'm encountering the same problem in the release version of AFDKO 2.5 build 65322 with a mark feature. So it seems to have crept into the pre-compiled release version somehow ...

typoman:
I'm encountering the same problem in the release version of AFDKO 2.5 build 65590. This does not occur in the the build 647800.

PR #50, opened 2015-05-05
Revert "Remove 32bit architecture dependence"

jamesgk:
Code in the FDK depends on the assumption that it will be compiled on a 32 bit platform. Until someone decides to actually remove all the 32bit-dependent code from the FDK, these flags should remain.

Fixes issue #48


behdad:
Right.  @miguelsousa, any chance you can merge this?  This currently makes afdko from github broken on 64bit architectures.

PR #51, opened 2015-05-09
[tx] make temporary files in the user's %TMP% folder on Windows

anthrotype:
as noted here (https://github.com/adobe-type-tools/afdko/issues/20), on Windows the stdio.h tmpfile function tries to create temporary files in the root folder, thus requiring adminstrative privileges.

I only patched the tx command, but I know the same issue affects all the afdko commands which use tmpfile.

Anyway, here is my workaround. Maybe somebody could help me find a better place for this code, so that it can be reused in other parts of the library...

thank you,

C.


readroberts:
Looks fine to me. Thanks!

PR #59, opened 2015-06-26
Fix libautohintlib path in Xcode project.

codeman38:
For some reason, the Xcode project for autohint was looking for libautohintlib using an explicit path to a DerivedData directory under ~/Library. This caused a build failure for anyone who didn't have the library compiled in that exact place.

This should hopefully fix that, by changing the build to find that library via its subproject.


anthrotype:
@readroberts could you merge this please? it fixes a bug introduced in https://github.com/adobe-type-tools/afdko/commit/6105e54cbe549c351d4b625141fa7843f35e4fd1

anthrotype:
thanks!

PR #71, opened 2015-09-28
Support hyphens in glyph names

schriftgestalt:
Allows hyphens in glyph names. It takes care of both, names with hyphens and ranges. 

And there is a commit that allows design names in feature files for CID fonts. It uses the GlyphOrderAndAliasDB file like for name keyed fonts.


benkiel:
Georg, you might want to take a look at this discussion about hyphens in glyph names in the fonttools project: https://github.com/behdad/fonttools/pull/334#issuecomment-130140529

schriftgestalt:
> I will change makeotf to allow characters which are not permitted in glyph names by the PostScript spec. Miguel Sousa and Frank Grießhammer discussed this with me, and I do see that there are advantages in using these in the development-mode glyph names. I am still reluctant to support hyphen because the FDK has been in use for over 10 years, and allowing a hyphen in glyph names may break existing feature files. Is there any reason not to use another character? I currently plan to allow "+ * : ~ ^ !".

I fixed the problem with the existing feature files so I don't see a problem.

miguelsousa:
@schriftgestalt I (and others on the team) think that this is a bad idea. The hyphen has been in use to denote glyph ranges in FEA syntax, and it's also used with the command line tools to denote glyph ranges. We recommend that you use one of these `+ * : ~ ^ !`.

PR #84, opened 2015-11-22
Fix broken ufonormalizer call

i-tu:
The `ufonormalizerLib` import name changed to `ufonormalizer` in commit 28c50d7.
Line 419 was changed accordingly, but the call on 436 was missed. This commit fixes that.


i-tu:
Glad it helped :)

PR #99, opened 2016-01-12
Allow glyph names with up to 63 characters

brawer:
readroberts wrote on Sep 25, 2015: "About the 31 character limit in
PostScript glyph names: although the spec clearly states the limit of
31 characters, I increased the limit in the AFDKO tools to 63 a year
or to ago for practical reasons. Use of the rules to name complex
ligatures leads to longer glyph names, and I had reasonable use cases
that would exceed this limit. A number of the limits stated in the
spec are not in fact enforced any the rasterizes I have available for
testing. I changed this limit in the AFDKO tools in order to test if
we could in practice get away with longer glyph names. Although there
are few such cases, I have not heard of any reports of difficulties. I
agree that the spec should be changed."

https://github.com/fontforge/fontforge/pull/2500#issuecomment-143263393


adrientetar:
@brawer: Can you also do the change for named glyph classes? (Section 2.g.ii.) 
My bug report was initially aimed at named classes, which makeotf also accepts longer ones.

brawer:
Sure, done.

PR #116, opened 2016-03-28
[ttxn] catch GetoptError raised by ttx.parseOptions when no input file is provided

### LINKED ISSUE #115: Bug in TTXN again (probably because of changes in TTX (FontTools)
There is a bug in ttxn.py again. It doesn't work with the latest FontTools. ttxn.py was fixed in January 2016. And if I replace ttxn.py in the official AFDKO 2.5 64958, ttxn from January 2016 works (more or less). (TTXN contained in AFDKO 2.5 64958 does not work.) But in the branch of the AFDKO and with the FontTools (both from 03-26-2016) ttxn does not work anymore. So either the "bug" has come back or it is a new "bug". Well, it seems to be more an incompatibility with FontTools.

I think, TTXN should be integrated in FontTools but not in the AFDKO, because I assume, that this reduces the risk for these incompatibilities.

anthrotype:
Should fix #115 

@Arno-Enslin please confirm.
Thanks

### DIFF
--- a/FDK/Tools/SharedData/FDKScripts/ttxn.py
+++ b/FDK/Tools/SharedData/FDKScripts/ttxn.py
@@ -91,6 +91,7 @@ import collections
 import textwrap
 import string
 import platform
+import getopt
 
 curSystem = platform.system()
 INDENT = "   "
@@ -2277,8 +2278,13 @@ def run(args):
 		supressTTFDiffs = 1
 	else:
 		supressTTFDiffs = 0
-		
-	jobs, options = ttx.parseOptions(args)
+
+	try:
+		jobs, options = ttx.parseOptions(args)
+	except getopt.GetoptError as e:
+		print "ERROR:", e
+		sys.exit(2)
+
 	if not jobs:
 		return
 		
--- a/FDK/Tools/SharedData/FDKScripts/ttxn.py
+++ b/FDK/Tools/SharedData/FDKScripts/ttxn.py
@@ -52,6 +52,7 @@ Get feat tag list. For each tag:
 					
 		
 """
+from __future__ import print_function, absolute_import
 
 __copyright__ = """Copyright 2014 Adobe Systems Incorporated (http://www.adobe.com/). All Rights Reserved.
 """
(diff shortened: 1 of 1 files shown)


Arno-Enslin:
@ anthrotype

No, I am getting still the same error.

The whole message:

```
Dumping "Decompile\SourceSerifPro-Regular.ttf" to "DecompileD\SourceSerifPro-Reg
ular.ttx"...
C:\FDK_26.03.2016\Tools\SharedData\FDKScripts\ttxn.py:1943: UserWarning: 'verbos
e' is deprecated; configure logging instead
  fontNumber)
  File "C:\FDK_26.03.2016\Tools\SharedData\FDKScripts\ttxn.py", line 2304, in <m
odule>
    run(sys.argv[1:])
  File "C:\FDK_26.03.2016\Tools\SharedData\FDKScripts\ttxn.py", line 2296, in ru
n
    ttf = ttnDump(input, output, options, showExtensionFlag, supressHints, supre
ssVersions, supressTTFDiffs)
  File "C:\FDK_26.03.2016\Tools\SharedData\FDKScripts\ttxn.py", line 2202, in tt
nDump
    disassembleInstructions=options.disassembleInstructions)
  File "C:\FDK_26.03.2016\Tools\win\Python\AFDKOPython27\lib\site-packages\fontt
ools-3.0-py2.7.egg\fontTools\ttLib\__init__.py", line 302, in saveXML
    self._tableToXML(tableWriter, tag, progress)
TypeError: _tableToXML() takes exactly 5 arguments (4 given)
```

anthrotype:
I pushed two more commits that should fix the issue reported by @Arno-Enslin and make the ttxn.py script work with latest behdad/fonttools master/HEAD.

The changes were more than I had expected, but that is also because ttxn.py code is quite repetitive. Several chunks could be factored out into functions, but since I was new it (I had never used it before), I didn't want to change too much and possibly break stuff.

Incidentally, this also makes ttxn.py compatible with both Python 2 and 3 (just like fonttools itself). I know this isn't a requirement as the AFDKO ships its own Python 2.7, but may become useful later.

@readroberts could you maybe have a look at the patch in your spare time?

BTW, I pinpointed a couple of lines where variables are assigned but never used. I wasn't sure what their purpose was, so I left them in there with a question mark (see comments starting with `# XXX`). This is a good chance to clean these up.

Arno-Enslin:
@ anthrotype

No errors anymore.

--------------------------

What's the best way to keep my local AFDKO with all modules/extensions always up to date? With regard to the dependencies. I would like to automatize that.

afdko-build-dir

Issue #134, opened 2016-04-22
Bug in StemHist.py: Two "-o options"


Arno-Enslin:
From the help of StemHist.py:

```
-o report path 
	[…]

-o  Don't print glyph names - just print a dot occasionally to show progress. 
```

I think, "Don't print glyph names" is the -q option.

I will create a pull request.

PR #135, opened 2016-04-22
Update StemHist.py

Arno-Enslin:
I had reported the issue ["there".](https://github.com/adobe-type-tools/afdko/issues/134)


readroberts:
Thanks! I do appreciate your going through all the issues carefully, UI as well as functional.

Arno-Enslin:
No need to thank me! Thank you and the other developers for your work. I have not searched for issues. I only have found them.

Maybe you can encourage potential users of the AFDKO [on this page](http://www.adobe.com/devnet/opentype/afdko.html), to build it from the Github source by themselves. I think, this would require build notes, that are a little bit more detailed with regard to the VC10 compiler in case of Windows. My advisement is, that this could raise the number of users, that report those bugs. I mean, it is conspicuous, that some of the heavier bugs were not reported yet.

PR #156, opened 2016-12-16
Fix help of rotateFont

mashabow:


readroberts:
Thanks for pointing out the errors, and especially the duplication of the entire help content. I didn't merge your PR as I would like to keep the example, but I copied all the other fixes from your PR.

mashabow:
Thank you!

PR #170, opened 2017-02-20
Improve output & wrapping of FinishInstall... scripts.

frankrolf:
The output of FinishInstallOSX was quite unreadable.
I wrapped every output to 70 characters, and made some minor optimizations.


miguelsousa:
@frankrolf FYI, I did a **rebase merging** instead of the usual plain merging. I think this makes the history tree cleaner.

PR #171, opened 2017-02-27
[mkicf.pl] Fix vertical baseline tags

mashabow:
See https://www.microsoft.com/typography/otspec/baselinetags.htm


kenlunde:
Looking back at my own records, I fixed these two tags in mid-2000, meaning almost 17 years ago! I have no clue how they managed to remain unfixed in AFDKO.

PR #199, opened 2017-10-13
Make otf2otc and otc2otf work with Python 2 and 3

moyogo:


miguelsousa:
Looks good. Thanks @moyogo

PR #201, opened 2017-10-25
Updating version number and change dates for otc2otf.py and otf2otc.py

cjchapman:
Just updating the version number and change dates to reflect the changes in moyogo & miguelsousa's recent commits.


moyogo:
Thanks!

PR #226, opened 2018-01-03
Improvements in CheckOutlinesUFO glyph list handling:

cjchapman:
Now using defcon_font.keys() if no glyph list specified, and defcon_font.glyphOrder if the glyph list is specified.
Also added an error message for when the glyph list is specified and the defcon_font.glyphOrder is empty or missing.


cjchapman:
I'm closing this, deleting the branch, and starting over with a new version branch off of Read's recent changes.

PR #230, opened 2018-01-05
CheckOutlinesUFO pens work

cjchapman:


setup.py
```diff
@@ -154,6 +154,7 @@ def run(self):
 		  'mutatorMath>=2.1.0', #2.1.0
 		  'ufolib>=2.1.1',
 		  'ufonormalizer>=0.3.2',
+		  'fontPens>=0.1.0.dev0'
```
anthrotype: only that `fontPens` is not on PyPI yet
https://pypi.python.org/pypi?%3Aaction=search&term=fontPens
so running setup.py install or pip install will fail

cjchapman: May I help with getting `fontPens` on PyPI?

anthrotype: thanks, I can do it.
tracking it here https://github.com/robofab-developers/fontPens/issues/7

cjchapman: Awesome. Thanks! Also, after thinking about these changes further, I started to wonder about the change in behavior, which lead to this issue:
https://github.com/robofab-developers/fontPens/issues/8

If it's not expected behavior, I'd be happy to help by working on this issue.

Issue #231, opened 2018-01-08
python3 plans


typemytype:
now there is a mixture of script that support both py2 as py3 others aren't upgraded yet

are there any plans todo so?

miguelsousa:
Yes, the plan is to upgrade most of them. I don't say "all" because some may get deprecated.

typemytype:
I just converted the ones RoboFont requires... this is already a big start!!

miguelsousa:
Indeed. Thanks!

PR #232, opened 2018-01-08
partially making in py3 ready

typemytype:
there are still lots of improvements to make:

with open(path, "wr") as f:
f.read / f.write

lots of open files warnings...

https://github.com/adobe-type-tools/afdko/issues/231

### DIFF
--- a/afdko/Tools/SharedData/FDKScripts/BezTools.py
+++ b/afdko/Tools/SharedData/FDKScripts/BezTools.py
@@ -1,3 +1,16 @@
+from __future__ import print_function, absolute_import
+import sys
+
+if sys.version_info > (3,):
+	long = int
+	longType = long
+try:
+	from types import FloatType, StringType, LongType
+except:
+	FloatType = float
+	StringType = str
+	LongType = int
+
 """
 BezTools.py v 1.13 July 11 2017
 
@@ -7,19 +20,17 @@ Used by AC and focus/CheckOutlines.
 __copyright__ = """Copyright 2014-2017 Adobe Systems Incorporated (http://www.adobe.com/). All Rights Reserved.
 """
 
-import sys
 import re
 import time
 import os
-import FDKUtils
-import ConvertFontToCID
+from . import FDKUtils
+from . import ConvertFontToCID
 from fontTools.misc.psCharStrings import T2OutlineExtractor, SimpleT2Decompiler
 from fontTools.pens.basePen import BasePen
-from types import FloatType, StringType, LongType
 debug = 0
 def debugMsg(*args):
 	if debug:
-		print args
+		print(args)
 kStackLimit = 46
 kStemLimit = 96
 
@@ -59,7 +70,7 @@ class T2ToBezExtractor(T2OutlineExtractor):
 		self.subrLevel -= 1
 		if (not self.closePathSeen) and (self.subrLevel == 0):
 			 self.closePath()
-			
+
 	def rMoveTo(self, point):
 		point = self._nextPoint(point)
 		if not self.firstMarkingOpSeen :
@@ -213,7 +224,7 @@ class T2ToBezExtractor(T2OutlineExtractor):
 
 		for i in range(numhhints/2):
 			if hintOn(i, hintMaskBytes):
-				curhhints.extend(self.hhints[2*i:2*i+2])	
+				curhhints.extend(self.hhints[2*i:2*i+2])
 		numvhints = len(self.vhints)
 		for i in range(numvhints/2):
 			if hintOn(i + numhhints/2, hintMaskBytes):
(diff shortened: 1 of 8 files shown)


afdko/Tools/SharedData/FDKScripts/BezTools.py
```diff
@@ -1,3 +1,16 @@
+from __future__ import print_function, absolute_import
+import sys
+
+if sys.version_info > (3,):
+	long = int
+	longType = long
```
anthrotype: why do you need both `longType` (lowercase initial) and `LongType` (uppercase), both alias of `int` on py3?

afdko/Tools/SharedData/FDKScripts/FDKUtils.py
```diff
@@ -57,22 +59,22 @@ def findFDKDirs():
...
-		print "Fatal error: could not find '%s or %s or %s'." % (path,p1,p2)
+		print("Fatal error: could not find '%s or %s or %s'." % (path,p1,p2))
 	raise FDKEnvError
 
 def runShellCmd(cmd):
 	try:
 		p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT).stdout
 		log = p.read()
-		return log
+		return str(log)
```
anthrotype: that's wrong. `log` returned by subprocess is a bytes string. On py3, if you do `str(b"a")` without an `encoding` argument, you get its repr `"b'a'". You want to call `decode` method of bytes with the encoding that the subprocess command is using to print to stdout (probably ascii or utf-8, but you have to know).

afdko/Tools/SharedData/FDKScripts/BezTools.py
typemytype: oh fixed 14446c2a79eaa5e1a5bf843c904e200f9515e167

afdko/Tools/SharedData/FDKScripts/MakeOTF.py
```diff
@@ -1388,7 +1394,7 @@ def getROS(fontPath):
 	"""
 	R=O=S=None
 	fp = open(fontPath, "rb")
-	data = fp.read(5000)
+	data = str(fp.read(5000))
```
anthrotype: again, wrong for the same reason I above. you're calling `str` on a bytes string, but it looks like you want to decode it.
Just open the file with `io.open` as text and pass the correct encoding. You'll read properly decoded (unicode) strings.

afdko/Tools/SharedData/FDKScripts/FDKUtils.py
typemytype: True, this is really wrong... but it works. To my defence: the whole file is a mess and needs a decent fix up...

will fix it

afdko/Tools/SharedData/FDKScripts/MakeOTF.py
typemytype: see above

afdko/Tools/SharedData/FDKScripts/autohint.py
```diff
@@ -13,7 +15,7 @@
 Auto-hinting program for PostScript and OpenType/CFF fonts.
 """
 
-from ufoTools import kProcessedGlyphsLayerName, kProcessedGlyphsLayer
+from .ufoTools import kProcessedGlyphsLayerName, kProcessedGlyphsLayer
```
anthrotype: careful with these relative imports. if the python module is meant to be run as a script, then you can't use relative imports.

anthrotype: if one attempts to run a python script containing a relative import, one gets
`ValueError: Attempted relative import in non-package` (py2) or `ModuleNotFoundError:... '__main__' is not a package` (py3)

typemytype: will revert

afdko/Tools/SharedData/FDKScripts/FDKUtils.py
anthrotype: agree

afdko/Tools/SharedData/FDKScripts/autohint.py
anthrotype: note that the issue with scripts and relative imports only arises if one attempts to call the module as a script like in `python afdko/Tools/SharedData/FDKScripts/autohint.py`.
If the the module has a setuptools controle_scripts entry point defined in setup.py, then it will be imported as a module and it's main() will be called from the wrapper script. So in this case, the name of the module is not `__main__` but the full qualified module name.

afdko/Tools/SharedData/FDKScripts/FDKUtils.py
typemytype: fixed ac1494d1c3ba2a0fd99b13634b8aae217552ba92

afdko/Tools/SharedData/FDKScripts/autohint.py
typemytype: reverted ac1494d1c3ba2a0fd99b13634b8aae217552ba92

PR #233, opened 2018-01-08
py3 improvements 

typemytype:


afdko/Tools/SharedData/FDKScripts/BezTools.py
```diff
@@ -1,9 +1,8 @@
-from __future__ import print_function, absolute_import
```
anthrotype: the `absolute_import` should stay. It simply makes the import machinery in python2 behave like python3 (e.g. disallowing implicit relative imports)

afdko/Tools/SharedData/FDKScripts/BezTools.py
```diff
@@ -23,8 +22,8 @@
 import re
 import time
 import os
-from . import FDKUtils
-from . import ConvertFontToCID
+import FDKUtils
```
anthrotype: this `import FDKUtils` will fail on python 3 if it is meant to be an implicit relative import: ie. the FDKUtils.py is a module that happens to sit in the same package directory as the current module which is importing it.
It only works if FDKUtils is an absolute import (because it doesn't have a ".").
I don't know what the packages layout in afdko is meant to be.
is the current module intented to be imported as `afdko.Tools.SharedData.FDKScripts.BezTools`?

afdko/Tools/SharedData/FDKScripts/BezTools.py
```diff
@@ -1298,7 +1297,7 @@ def getFontInfo(self, fontPSName, inputPath, allow_no_blues, noFlex, vCounterGly
 		for i in range(3, numBlueValues,2):
 			blueValues[i] = blueValues[i] - blueValues[i-1]
 
-		blueValues = map(str, blueValues)
+		blueValues = list(map(str, blueValues))
```
anthrotype: `[str(v) for v in blueValues]` is more idiomatic

typemytype: see c559f33b4b1c3aa8f3768eb15159edec5ea55134

afdko/Tools/SharedData/FDKScripts/BezTools.py
typemytype: if I run `python3 autohint` (that import bezTools and FDKUtils) it works fine

fdk is a script collection, not a package, how ever sometimes they need each other. Its not clear :)

anthrotype: exactly, it works if you run `python3 autohint.py` from the same directory (the current working directory is always in the PYTHONPATH), in which case absolute or relative imports are the same thing.

typemytype: I did both `python3 full/path/to/autohint.py` as what FDK is actually doing cd'ing to the tools folder and execute from there, see https://github.com/adobe-type-tools/afdko/blob/master/afdko/Tools/osx/autohint

anthrotype: I haven't looked into it, but I think the `setFDKPaths` shell script must be doing something to ensure that the FDKScripts directory is in the PYTHONPATH. In which case `import FDKUtils` does work.

khaledhosny:
Note that I already make autohint Py23 compatible over in https://github.com/khaledhosny/psautohint, just in case anyone is interested in taking a look.

miguelsousa:
@typemytype @anthrotype Is this good to merge now?

anthrotype:
> Is this good to merge now?

I haven't tested. If Frederik is using it already you may as well merge.

Issue #237, opened 2018-01-11
'cvParameters' block is allowed only Character Variant (cvXX) features.


miguelsousa:
After applying the workaround in #236 attempting to build [Source Code Regular](https://github.com/adobe-fonts/source-code-pro/tree/master/Roman/Instances/Regular) fails with
```
makeotfexe [FATAL] <SourceCodePro-Regular> A 'cvParameters' block is allowed 
only Character Variant (cvXX) features. It is being used in feature 'cv10'.
```

The code for the `cv10` feature is [here](https://github.com/adobe-fonts/source-code-pro/blob/master/Roman/familyGSUB.fea#L469-L480) and it's structured like the previous cvXX features.

cjchapman:
I've got a proposed fix for this here:
https://github.com/adobe-type-tools/afdko/pull/243

miguelsousa:
Fixed by #243. afdko v2.6.25

PR #243, opened 2018-01-17
Fix for issue #237:

cjchapman:
multiply first digit of cv number by 10 instead of shifting left by 8 bits, so now we get a cv number of 10 instead of 256 for cv10.


cjchapman:
I forgot the "#" in my original commit, so let me explicitly link the issue here:
https://github.com/adobe-type-tools/afdko/issues/237

cjchapman:
I'm merging this per conversation with Miguel.

PR #252, opened 2018-01-30
Remove Perl scripts

miguelsousa:
These are superseded by the scripts at https://github.com/adobe-type-tools/perl-scripts
Referenced in https://github.com/adobe-type-tools/perl-scripts/issues/2


miguelsousa:
@kenlunde no worries. Thanks

PR #255, opened 2018-02-01
Run flake8 on cleaned up scripts

miguelsousa:


miguelsousa:
@cjchapman the Codecov bot post is a bit of an overkill at this point.

cjchapman:
@miguelsousa I turned off the Codecov comments and also the status check for now.

PR #266, opened 2018-02-06
Fixed assert "out of range region index found in item variation store…

blueshade7:
… subtable"

Fixed an assert "out of range region index found in item variation store subtable " in var_getIVSRegionIndices because a wrong region count was used to range-check region indexes.
The bug was exposed when reading Monotype AvenirNext-Variable.ttf

### DIFF
--- a/afdko/Tools/Programs/public/lib/source/varread/varread.c
+++ b/afdko/Tools/Programs/public/lib/source/varread/varread.c
@@ -998,7 +998,7 @@ static float var_applyDeltasForIndexPair(ctlSharedStmCallbacks *sscb, var_itemVa
         return netAdjustment;
     }
 
-    subRegionCount = var_getIVSRegionIndices(ivs, pair->outerIndex, regionIndices, subtable->regionCount);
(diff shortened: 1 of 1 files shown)


readroberts:
How did this cause a bug? It looks to me as if the regionCount is either the subtable region count, or the IVS region list count, and is used only to make sure that no region index exceeds this value. Also, I think it might be useful to make sure regionCount is set the same in all the call chains to var_getIVSRegionIndices.

blueshade7:
The value of variable regionCount is taken from the variation region list, while subtable->regionCount is the number of region indices referenced by a subtable, which may be less than regionCount. This field in itemVariationDataSubtable struct is better to be renamed to "regionIndexCount" to avoid confusion actually, the name used in OT spec for this data.

When a smaller value (subtable->regionCount) than actual (regionCount) is passed to var_getIVSRegionIndices, it may reject a valid region index as invalid.

readroberts:
I will open a new issue for the variable name confusion.

miguelsousa:
Follow up issue #272

PR #248, opened 2018-01-29
Clean up lint errors in several Python scripts

miguelsousa:

PR #270, opened 2018-02-08
Undo None removals performed at #248

miguelsousa:


miguelsousa:
@cjchapman thanks for catching the line 755. The change at line 922 I decided to leave as is because I'm sure it's not going to cause any problems.

PR #285, opened 2018-03-19
first pass at .appveyor.yml

cjchapman:


.appveyor.yml
```diff
@@ -0,0 +1,14 @@
+install:
+  - python -m pip install --upgrade pip
+  - pip install wheel
+  - python setup.py bdist_wheel
+  - dir
+  - dir dist
+  - dir dist\*.whl
+  - pip install dist\*.whl
```
anthrotype: try doing `pip install --no-index --find-links dist afdko`

.appveyor.yml
```diff
@@ -0,0 +1,14 @@
+install:
+  - python -m pip install --upgrade pip
+  - pip install wheel
+  - python setup.py bdist_wheel
+  - dir
+  - dir dist
+  - dir dist\*.whl
+  - pip install .\dist\*.whl
```
anthrotype: I believe pip simple hates backslashes

cjchapman: @anthrotype Just saw your comment after my last commit, thanks!

anthrotype: I think the problem was with the `*` not being expanded by cmd.exe for some reasons. I see that it works when you `dir`.
Anyway, the `--find-links <local_dir>` option is better than globbing for wheels

anthrotype: oh. sorry I forgot.. you also need to pass `--no-deps`, otherwise it attempts to install the install requirements from the local directory (since we forced `--no-index`)

anthrotype: for the other dependencies, you could add a requirements.txt file (ideally with pinned-down versions so the test runs are deterministic/reproducible) which you `pip install -r` from.

or maybe for now simply do `for %x in (dir dist\*.whl) do python -m pip install %x`

cjchapman: I think --no-deps means it won't install the dependencies, but I do want it to install the dependencies.

.appveyor.yml
```diff
@@ -2,10 +2,7 @@ install:
   - python -m pip install --upgrade pip
   - pip install wheel
   - python setup.py bdist_wheel
-  - dir
-  - dir dist
-  - dir dist\*.whl
-  - pip install --no-index --find-links dist afdko
+  - for %%f in (dist\*.whl) do pip install %%f
```
anthrotype: exactly :+1:

PR #289, opened 2018-03-28
make Travis deploy dev releases to PyPI

cjchapman:


.travis.yml
```diff
@@ -49,3 +49,11 @@ notifications:
...
     on_success: change
     on_failure: always
+deploy:
+  provider: pypi
+  user: afdko-travis
+  password:
+    secure: fkjaRurDm3ccb9IUPqDNd5VX1PkZnPrraedHigbpRmc11e0xn+AwWEBuI9EhFLlIl3omzxmajw4UXQb2qw0pIyLPUGnASZvsHWFE9mUz6v2XHYSI1RLG+xpyd0F7ZGExAW6U5fca1uTUTIo68/CJCMk0NSrcVGi6WkaFOXsQ+45UX7uSfCvPsA80w1DnA3tgrrbaLPaz2y9IZmdmMU8DQzE5IU8ODtXS1exQV1YUhlapRvvOkmhv61edVlTeqzd9dCQPWHxegqs6hkiUVX01H0wYNQkaomY6K1O8AmOmDvacXRQILkB/H5luQNGtOhRMGIwr65VIjLl2xNVgEzVAT3rYQ31oz794AkpjDGvzccNgUT9BjwVjxbBrYqpltfYs6q8HlQ8etl1GMjMXrHoZLdNlemKsyAghpNOrQD9pTHYM8OU7m6tCxSDzLnMwnGpjHohsfDg6WWfoBybyAK2RwJ0/Okv/lgdPdupaPOmsJb8h0ymuQB7StAoWYAql2VO+JrFOLlBar+n/2LtjVvW6KISpDSE7qrJEtEfO0ElU8Pc/T+cE5qBlBxa3wh87nqBPee/oOcJNuNmHCk7YzWCXShj+fiDjPQYy6WLHZ1/e8I1AV3DEWlqOAnGja46FARRHLZ8GNLhZ8DAiRsgV4rd9NCl3v7UJWp036Tr7KAZP2nc=
+  on:
+    distributions: bdist_wheel
+    repo: adobe-type-tools/afdko
```
anthrotype: Would this mean you are uploading wheels to pypi at _every_ commit to master?
I believe that would be a bit of a waste of resources.

miguelsousa: Not at every commit, but at every time `master` changes via a PR. Our goal is to do something similar to [pybuilder](https://pypi.python.org/pypi/pybuilder), i.e. always have dev wheels that reflect the state of `master`, and featured releases that are issued when `master` is tagged.

anthrotype: I see you added `branches: only: master` with https://github.com/adobe-type-tools/afdko/pull/291, but that simply tell Travis to exclude non master branches from the build. It doesn't affect the deploy step.

The `on:` section of `deploy` is what makes the latter conditional. Your only condition is `repo:`
https://docs.travis-ci.com/user/deployment#Conditional-Releases-with-on%3A

also, `distributions:` should not be indented under `on:`, it's indepedent from it:
https://docs.travis-ci.com/user/deployment/pypi/#Uploading-different-distributions

cjchapman: @anthrotype Sorry for my delayed responses -- I was on leave the past few days. My understanding that `branches: only: master` prevents pushes to other branches from building (in general), and allows pushes to master to build (and deploy), which is what we want, while PRs from any branch to master still build regardless of this (and Travis behavior with recent PRs seems consistent with this understanding). It's also my understanding that PR builds will not deploy, as the Travis PyPI documentation says:
>Builds triggered from Pull Requests will never trigger a release.

Also, we get this message at the end of PR builds:

>Skipping a deployment with the pypi provider because the current build is a pull request.

If you can suggest a cleaner way to do dev releases to PyPI of PRs that we merge to master, I'm definitely open to alternatives. :-)

anthrotype: > allows pushes to master to build (and deploy), which is what we want

yes, that's what I personally find a bit excessive: i.e. uploading a (dev) wheel every time Travis runs a build from the master branch.

The PyPI Warehouse project has historically struggled for funding and it's hosted on servers provided for free by a few sponsors. I'm not affiliated to them, and I don't mean to be judging what is "fair usage" or not. I just think that if you're going to upload very often in between releases, then it may make sense to host these development wheels elsewhere, on some Adobe server maybe.
It's not difficult to set up a PyPI index.

Anyway, it seems the deploy stage is currently failing for you (even though Travis doesn't exist with non-zero, but that's a different Travis bug https://github.com/travis-ci/travis-ci/issues/1574)

see https://travis-ci.org/adobe-type-tools/afdko/jobs/361911643#L5432
```
sh: 1: twine: not found
```

I think this is because you are sourcing the virtual environment in the current Travis shell, and later on when Travis own deploy script installs twine in the global python, the venv's environment appear to be masking that. In fonttools we run all the ci commands in sub-shells using some bash scripts, so the Travis global environment isn't affected.

anthrotype: or it might have to do with this 
https://github.com/adobe-type-tools/afdko/commit/e3a913c5139003a3a9d46ed77d7baec2b5c15540

what's that about?

anthrotype: I don't have those issues with Travis PyPI depoloyment mentioned in https://github.com/travis-ci/travis-ci/issues/9303 -- but I always use `sudo: false` (i.e. containerized) Travis images in my projects as I don't usually need to sudo apt get stuff.

cjchapman: @miguelsousa in light of @anthrotype's suggestions, I propose that we switch to the more common model of deploying based on tags (rather than every merge to master) and then have some policy of tagging dev releases fairly regularly (perhaps monthly?), and perhaps also deleting older dev releases so as not to waste too much server disk space.

PR #298, opened 2018-04-04
[checkoutlinesufo] Replace doctest tests with pytest tests

miguelsousa:


.appveyor.yml
```diff
@@ -1,12 +1,14 @@
...
   - for %%f in (dist\*.whl) do pip install %%f
 
 build: false
 
 test_script:
-  - cd afdko\Tools\SharedData\FDKScripts
-  - python CheckOutlinesUFO.py --test
+  - cd Tests
+  - python checkoutlinesufo_test.py
+  - python ttfcomponentizer_test.py
```
anthrotype: Why don’t you call just`pytest` and let it discover the `*_test.py` modules like it is good at doing?

miguelsousa: Because I didn't know it could do that :grin:

miguelsousa: Can I just call `pytest` without being inside the Tests folder?

cjchapman: @miguelsousa in addition to the change @anthrotype suggested above, I think we can do something similar using [pytest-cov](https://pypi.python.org/pypi/pytest-cov) in `.travis.yml`. I'm going to go ahead and approve this commit now, and then make the pytest related changes in a new PR.

PR #307, opened 2018-04-07
Move wheel-making and -installation to later in the process

miguelsousa:
This way there's less scrolling to get to failing tests, and the coverage is reported for the source files, not for the installed ones.


miguelsousa:
https://codecov.io/gh/adobe-type-tools/afdko/pull/307

PR #314, opened 2018-04-18
Attempt to fix PyPI deployment

miguelsousa:


benkiel:
You will want to be sure that the pass function is in setup.py

.travis.yml
```diff
@@ -72,21 +72,25 @@ branches:
...
-    repo: adobe-type-tools/afdko
-    condition: "$BUILD_DIST == true"
+  # deploy to PyPI on tags
+  - provider: pypi
+    server: https://upload.pypi.org/legacy/
+    user: afdko-travis
+    password:
+      secure: fkjaRurDm3ccb9IUPqDNd5VX1PkZnPrraedHigbpRmc11e0xn+AwWEBuI9EhFLlIl3omzxmajw4UXQb2qw0pIyLPUGnASZvsHWFE9mUz6v2XHYSI1RLG+xpyd0F7ZGExAW6U5fca1uTUTIo68/CJCMk0NSrcVGi6WkaFOXsQ+45UX7uSfCvPsA80w1DnA3tgrrbaLPaz2y9IZmdmMU8DQzE5IU8ODtXS1exQV1YUhlapRvvOkmhv61edVlTeqzd9dCQPWHxegqs6hkiUVX01H0wYNQkaomY6K1O8AmOmDvacXRQILkB/H5luQNGtOhRMGIwr65VIjLl2xNVgEzVAT3rYQ31oz794AkpjDGvzccNgUT9BjwVjxbBrYqpltfYs6q8HlQ8etl1GMjMXrHoZLdNlemKsyAghpNOrQD9pTHYM8OU7m6tCxSDzLnMwnGpjHohsfDg6WWfoBybyAK2RwJ0/Okv/lgdPdupaPOmsJb8h0ymuQB7StAoWYAql2VO+JrFOLlBar+n/2LtjVvW6KISpDSE7qrJEtEfO0ElU8Pc/T+cE5qBlBxa3wh87nqBPee/oOcJNuNmHCk7YzWCXShj+fiDjPQYy6WLHZ1/e8I1AV3DEWlqOAnGja46FARRHLZ8GNLhZ8DAiRsgV4rd9NCl3v7UJWp036Tr7KAZP2nc=
+    skip_cleanup: true
+    distributions: pass
```
anthrotype: yeah, that "pass" thing is simply a way to trick the Travis dpl tool to run `python setup.py pass` and in fact _not_ do anything. The default is that it would do `python setup.py sdist bdist_wheel`. But if you have already built your distribution packages (e.g. for C extension modules we want to compile them inside a manylinyux1 docker container for greater portability), we don't want Travis to attempt to rebuilt them. If we `skip_cleanup` and place the sdist and wheel packages inside the default `dist/` folder, then the Travis dpl tool will find them and upload those instead of rebuilding them again as part of the deploy stage.

Lately, instead of defining a custom setuptools "pass" command that does nothing, I've been passing the built-in `check` command to the deploy `distributions:` to achieve the same effect. Doing `setup.py check` does basic metadata checks (may return non-zero if something is not ok), but does not build anything.

miguelsousa: I must have done something wrong because AFAICT Travis is trying to build them.

miguelsousa:
Still no go https://travis-ci.org/adobe-type-tools/afdko/builds/368120714
This part in particular doesn't make sense to me,
```sh
Successfully installed pip-10.0.0
./.travis/deploy.sh: line 5: pip: command not found
```

miguelsousa:
Is it because `pip` was installed with `--user`?

miguelsousa:
Success at last https://pypi.org/project/afdko/2.6.26.dev20180418133217/  😌

PR #315, opened 2018-04-18
use git to compute version and files for source distribution 

anthrotype:
With [setuptools_scm](https://github.com/pypa/setuptools_scm), you can stop worrying about the package version and the list of files to include in the source distribution, as that's all done automatically for you by your version control system.

When you want to push a release, just push a git tag. That will become the exact package version.
When you generate a sdist, all the files that are tracked by git are automatically included.

Currently attempting to install from the sdist is broken, as there are many source files missing. And in fact the canonical sdist package is missing from PyPI, only a mac wheel is present. Which means I can't pip install afdko on my Linux machine, even if I do have all the required compiler toolchain (sure, I can clone and install from there, but that's not the point I'm trying to make here).

Maintaining the MANIFEST.in manually is error prone, and is a waste of time, since git already knows what are the source files.

One caveat. Because of the special way setuptools_scm treats the `.dev` suffix in a packge version string -- where it means the number of commits since the latest tag --, since there is already a tag that contains `.dev` (i.e. the one Miguel pushed early on), that's going to produce an error when activating setuptools_scm plugin (you'll see from the Travis CI log). 

To make the error go away, you need to delete the tag containing the `.dev` suffix from both the local and remote repositories.

```sh
git tag -d <name_of_the_tag>   # removes it from the local repo
git push origin :refs/tags/<name_of_the_tag>  # deletes it from the remote repo
```

This explains why you can't use `.dev` in your tags if you wish to use setuptools_scm: https://github.com/pypa/setuptools_scm/issues/213

In order to tag pre-release tags, you can still use `2.27.0a1` or `2.27.0b3` or `2.27.0rc2` (alpha, beta, release candidate, with a progressive number, these are all PEP440 compliant version strings).


anthrotype:
https://ci.appveyor.com/project/adobe-type-tools/afdko/build/1.0.74#L60
```
AssertionError: own dev numbers are unsupported
```

yep. setup is gonna fail until you delete the tag as I explained above.

miguelsousa:
Just removed the recent `.dev` tag. Reviewing the changes now. Thanks.

anthrotype:
the travis failure is related to the fact that the system python on macos has old SSL library and the new pypi enforces some new protocol, and setuptools is attempting to pull the setup_requires when doing the `setup.py checkdocs`

I believe upgrading pip and setuptools should fix it

miguelsousa:
Is it just `flake8 setup.py` that is making it fail now?

anthrotype:
ok I think I'm done for now. this is ready for you to review

miguelsousa:
Thanks Cosimo. Getting all the planets to align is maddening.

cjchapman:
@anthrotype Let me also say: Thank you Cosimo for all your help!

anthrotype:
yeah..  the nice thing about the new `pyproject.toml` is that allows pip to bootstrap the build requirements before the installation from a source distribution. This meas that, provided one has the latest pip 10 or greater, one can install from the sdist and, in this case, the latest setuptools, setuptools_scm and wheel packages are going to be installed automatically in a temporary builld environment where the build happens, so the user only needs to worry about doing `pip install afdko`.

Of course also provided all the planets align :sweat_smile:

anthrotype:
So hopefully the next time you tag, Travis should build and upload both a sdist and a (mac) wheel to PyPI. :crossed_fingers:

miguelsousa:
Well, gonna try it right now.

anthrotype:
it worked! :tada: 

https://pypi.org/project/afdko/#files

anthrotype:
I just did `pip install --pre afdko` from my linux box, and it just works :tm:

PR #316, opened 2018-04-18
appveyor: deploy windows wheel on tagged commits

anthrotype:
you should replace the pypi credentials and encrypt the password by going to 'Account → Encrypt data' tool on Appveyor website

see https://www.appveyor.com/docs/build-configuration/#secure-variables

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,9 +1,7 @@
 environment:
-  # XXX replace pypi credentials and encrypt password:
-  # https://www.appveyor.com/docs/build-configuration/#secure-variables
-  TWINE_USERNAME: "anthrotype"
+  TWINE_USERNAME: "afdko-travis"
   TWINE_PASSWORD:
-    secure: 3MYRWT1R2pwRunaCYHAqlQ==
+    secure: myMd6E3niwjGvxEHeChOlg==
 
 install:
   - python -m pip install --upgrade pip setuptools
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,3 +1,10 @@
+environment:
+  # XXX replace pypi credentials and encrypt password:
+  # https://www.appveyor.com/docs/build-configuration/#secure-variables
+  TWINE_USERNAME: "anthrotype"
+  TWINE_PASSWORD:
+    secure: 3MYRWT1R2pwRunaCYHAqlQ==
+
 install:
   - python -m pip install --upgrade pip setuptools
   - pip install setuptools_scm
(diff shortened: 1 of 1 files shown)


anthrotype:
appveyor also provides the built artifacts for each build, which is nice
https://ci.appveyor.com/project/adobe-type-tools/afdko/build/1.0.83/artifacts

anthrotype:
travis linux build failed for random unrelated network issues, try reloading the build and it should fix itself

miguelsousa:
Just restarted Travis. Will look into Appveyor once I get to my desk.

miguelsousa:
@anthrotype after merging I added a new tag `2.27.0a2`, but the Windows deployment didn't happen. Not sure where to look for the error, or if there's something misconfigured in AppVeyor's settings.

anthrotype:
the latest build has:

https://ci.appveyor.com/project/adobe-type-tools/afdko#L10027

> Not deploying as this is not a tagged commit

also in the https://ci.appveyor.com/project/adobe-type-tools/afdko/history, all the builds seems to be from master branch and no tags are shown.

I could be some settings that says only build from master or similar.

I have this in my projects' appveyor settings

screenshot 2018-04-18 22 07 19

anthrotype:
i remember with @cjchapman we realised that Travis CI was interpreting git tags as if they were different branches, so when only enabling the building from master, we were also excluding the tag "branches", so we had to change the travis settings to also allow building from branches that matches a pattern (`\d+\.\d+\.\d+.*`)

miguelsousa:
Branches to build is set to "All branches" as well.
What about in **Deployment**? Did you set any deployment provider?

anthrotype:
maybe try use the same regexp that you use for travis

screenshot 2018-04-18 22 12 45

https://www.appveyor.com/docs/branches/

anthrotype:
no, I don't set up any deployment setting from the appveyor website.

miguelsousa:
I've specified the branches to build, and replaced the tag on the last commit (https://github.com/adobe-type-tools/afdko/tags). Nothing seems to have happened. Perhaps need to make a new commit.

anthrotype:
> Branches to build is set to "All branches" as well.

hm.. you sure? this is weird. It seems to only build master, there are no other branches nor tags listed in the build history.

> All branches are built by default. [...]
> By default AppVeyor starts a new build on any push to GitHub, whether it’s a regular commit or a new tag.

anthrotype:
> Perhaps need to make a new commit.

totally, yes.

git commit --allow-empty -m "Something"

miguelsousa:
Done 🤞

anthrotype:
try adding this to .appveyor.yml (same as the one you have in .travis.yml)

```yaml
branches:
  only:
    - master
    - /^\d+\.\d+\.\d+.*$/
```

PR #318, opened 2018-04-18
appveyor: try add branches regexp to match tag names

anthrotype:


anthrotype:
hmm...

https://ci.appveyor.com/project/adobe-type-tools/afdko/build/1.0.85

> Error parsing appveyor.yml: (Line: 29, Col: 1, Idx: 549) - (Line: 29, Col: 9, Idx: 557): Duplicate key

but I don't see "branches" key anywhere else?!

anthrotype:
something weird is going on..
why didn't your empty "Blah" commit show up in the appveyor build history?
whereas the PRs are showing in the list?

and why do i get this "error parsing appveyor.yml"? I don't see what's wrong with the lines I added

miguelsousa:
@anthrotype I also don't know what's going on. Perhaps AppVeyor needs a break.
Yep, two commits I made went thru Travis but not AppVeyor somehow.

anthrotype:
what I'd do is completely nuke the appveyor settings and start from scratch.
deleting the weekhook from the github repo settings, and also deleting the appveyor page for the afdko and re-autorizing appveyor service on the repository again, with clean slate.
just leave the default settings (only change to look for .appveyor.yml with the leading dot, like the current file, as by default it looks for appveyor.yml without leading dot).

then it must work.

anthrotype:
because the docs are clear and say appveyor build all branches and all tags by default, so there must be some weird settings that's sticking somewhere. I'd clear everything and make sure the default settings are restored

miguelsousa:
@anthrotype I've reset AppVeyor as you suggested and all is working now, it seems. There's a wheel for Windows on PyPI https://pypi.org/project/afdko/#files
Going to close this PR.

Issue #251, opened 2018-01-29
Create & upload Linux package to PyPI


cjchapman:

miguelsousa:
Partially completed by #286

miguelsousa:
Finished via #315.

anthrotype:
technically we have only uploaded the source distribution package, which can be pip installed on linux (as well as on mac or windows) provided one has the compiler toolchain installed on the system.
It'd still be nice to also have a manylinux1 binary wheel package that can be installed on most Linux distros without needing to compile from source.

miguelsousa:
Oh, thanks for the explanation. [Your comment on #315](https://github.com/adobe-type-tools/afdko/pull/315#issuecomment-382479615) gave me the impression that this was done.

PR #319, opened 2018-04-19
build manylinux1 wheel for Linux python2.7

anthrotype:
fixes https://github.com/adobe-type-tools/afdko/issues/251

### DIFF
--- a/.travis/before_deploy.sh
+++ b/.travis/before_deploy.sh
@@ -7,8 +7,7 @@ set -x
 # Travis runs the `before_deploy` stage before each deployment, but
 # we only want to build them once, as we want to use the same
 # files for both Github and PyPI
-if $(ls ./dist/afdko*.zip > /dev/null 2>&1) && \
-		$(ls ./dist/afdko*.whl > /dev/null 2>&1); then
+if $(ls ./dist/afdko*.whl > /dev/null 2>&1); then
 	echo "Distribution packages already exists; skipping"
 else
 	if [ "$TRAVIS_OS_NAME" == "linux" ]; then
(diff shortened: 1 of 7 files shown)


anthrotype:
Looks like this additional "Linux-deploy" build that I added to the Travis matrix does not run from a PR, but only from pushes to master (i.e. once this PR branch is merged), since I configured the same as the other "OSX-deploy" build (`if: type == push`).

Anyway, I tested running the build_manylinux_wheel.sh script locally with the same docker image and seems to work. I get a afdko-...-cp27-cp27mu-manylinux1_x86_64.whl in the dist/ folder, ready to be uploaded with twine.

This proposed setup basically will run the PR builds from the default ubuntu Travis image, but will build the linux wheel for deployment from a PyPA manylinux1 docker container.

Maybe it would make more sense to build the manylinux wheel everytime, not just in the before_deploy step, and install that in the ubuntu Travis python (to verify it was built correctly) and run the tests against that?

.travis/before_deploy.sh
```diff
@@ -7,8 +7,7 @@ set -x
 # Travis runs the `before_deploy` stage before each deployment, but
 # we only want to build them once, as we want to use the same
 # files for both Github and PyPI
-if $(ls ./dist/afdko*.zip > /dev/null 2>&1) && \
```
anthrotype: (btw, the check for .zip was wrong, since we are using the default .tar.gz format for the sdist)

miguelsousa:
> Maybe it would make more sense to build the manylinux wheel everytime, not just in the before_deploy step, and install that in the ubuntu Travis python (to verify it was built correctly) and run the tests against that?

Sounds good to me.

miguelsousa:
@anthrotype I think something didn't work, because there's no additional wheel associated with `2.7.0b2` https://pypi.org/project/afdko/2.7.0b2/#files

miguelsousa:
oh, here's the error https://travis-ci.org/adobe-type-tools/afdko/jobs/368827801

anthrotype:
It thinks it is ruby.. probably missing some global language setting. I’ll look into it tomorrow

PR #320, opened 2018-04-19
[checkOutlinesUFO] Re-write the in-line help formatter so that it is …

readroberts:
…not a direct copy of the StackOverflow example

The StackOverflow license is viral. We can use code snippets as examples, but cannot directly copy code. We most certainly cannot include complete URLs that reference a StackOverflow example.

### DIFF
--- a/afdko/Tools/SharedData/FDKScripts/CheckOutlinesUFO.py
+++ b/afdko/Tools/SharedData/FDKScripts/CheckOutlinesUFO.py
@@ -256,35 +256,30 @@ def parse_glyph_list_arg(glyph_string):
     return glyph_list
 
 
-class CustomHelpFormatter(argparse.RawDescriptionHelpFormatter):
-    # This class is based on PreserveWhiteSpaceWrapRawTextHelpFormatter
-    # from: https://stackoverflow.com/questions/35917547/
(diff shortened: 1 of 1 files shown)


miguelsousa:
@readroberts is there a particular reason why you moved the `add_whitespace` method outside of the class?

readroberts:
@msousa @cjchapman I moved the 'add_whitespace' from a class method to a function because that is one of the flake8 complaints. Since 'add_whitespace' code never references 'self', and is not an inherited method, it does not need to be a method.

There is an outstanding Codacy warning that I don't know how to fix. The warning is that the arguments to the '_split_lines' method do not match the arguments to the inherited method. However, they do in fact match the arguments to the  parent class 'argparse.RawDescriptionHelpFormatter'. Chris, do you have any ideas about this?

miguelsousa:
> Since 'add_whitespace' code never references 'self', and is not an inherited method, it does not need to be a method.

That can be fixed by placing `@staticmethod` above the class method.

cjchapman:
@readroberts & @miguelsousa I can confirm that the super method for `_splint_lines` has the same arguments:
```python
class HelpFormatter(object):
...
    def _split_lines(self, text, width):
```
...so I recommend that we just ignore the spurious error for now, and disable it if we get many more false positives like this.

PR #325, opened 2018-04-20
travis: set 'language: python' else Travis thinks this is Ruby...

anthrotype:
https://travis-ci.org/adobe-type-tools/afdko/jobs/368822577#L474

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,7 @@ matrix:
         - DOCKER_IMAGE=quay.io/pypa/manylinux1_x86_64
         - BUILD_DIST=true
       if: type = push
+      language: python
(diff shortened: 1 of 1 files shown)


anthrotype:
Still failing.. this time the log seems truncated, can’t see any errors. Try click rebuild?

cjchapman:
I'm getting a timeout when I try to connect to Travis right now.

cjchapman:
OK, finally connected. Just hit restart. (I'm using my cellular connection on a train, which might be part of the problem.)

cjchapman:
@anthrotype Looks like it failed again, and again the log seems truncated, even the raw log.

PR #326, opened 2018-04-20
travis: skip the default Travis python 'install' step

anthrotype:
when no 'install:' is defined, and language is 'python', Travis automatically does
pip install -r requirements.txt

it shouldn't do that.

(i'm not sure this is going to change anything, but let's try anyway.. also, it's quite annoying that in order to test this, we have to merge a PR and tag all the time... though I see no other way)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,8 @@ matrix:
       if: type = push
       language: python
       python: 2.7
+      # skip default installation step
+      install: true
 
     - os: osx
       env:


cjchapman:
@anthrotype If I remember correctly, the reason I didn't specify python was that it already had python and when I specified python it failed on install. I'm not sure I remember correctly though. That was a million `.travis.yml` iterations ago...

PR #327, opened 2018-04-20
travis: also skip 'script' step, all the work happens in before_deploy

anthrotype:
not sure why for the other osx-deploy build, it doesn't complain that there is no 'script' step defined...
anyway, let's try one more time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,8 +66,10 @@ matrix:
       if: type = push
       language: python
       python: 2.7
-      # skip default installation step
+      # skip default 'install' and 'script' steps: all the work is in
+      # before_deploy
       install: true
+      script: true
 
     - os: osx
       env:


anthrotype:
so far so good.. now let's try to push a tag and see what happens.

miguelsousa:
I've set a tag `2.7.0b3`

anthrotype:
phew, I worked! :tada: 

https://pypi.org/project/afdko/2.7.0b3/#files

thanks for inviting me as collaborator :)

cjchapman:
@anthrotype Sorry about the delay with pushing a tag. Until @miguelsousa clarified, I thought you could (and would) push a tag. Thanks again for all your help!

miguelsousa:
Thank you. You've earned it. Hopefully it will make it easy for you when submitting patches.

PR #340, opened 2018-04-26
[appveyor] Build 32 and 64-bit wheels

miguelsousa:


anthrotype:
ideally, the afdko build system would build 32 bit binaries under a 32-bit python, and 64-bit binaries for 64-bit python (and "fat" 32+64 binaries on macOS). Currently it builds a mix of 32 or 64 bit binaries, and the generated wheel tags do not reflect the bitness of the native executables that are bundled with it.
The setup.py script could get the size of pointer from the current python interpreter, e.g. `import struct; bits = struct.calcsize('P') * 8)`, and then pass the right compiler flags to the makefiles.

readroberts:
I would argue against adding this complexity, for several reasons. 1) There is no need to do this, as in FDK all the binaries are stand-alone C programs, and there are no compiled Python modules. 2) As soon as we are totally happy with psautohint, it will replace autohint in the AFDKO, and then all the programs will be 64 bit, and 3) Apple announced in January that with 10.13.4, the OS will warn that 32-bit support is going away after High Sierra.

PR #341, opened 2018-04-26
[travis] Specify osx_image

miguelsousa:
Available versions at https://docs.travis-ci.com/user/reference/osx/#OS-X-Version

This is to force the wheel to support macOS 10.10 and above; current support is 10.12+

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ matrix:
         - pip uninstall --yes afdko
 
     - os: osx
+      osx_image: xcode6.4
       env: NAME=OSX
       if: type = pull_request
       language: generic


miguelsousa:
I checked the log and the wheel's name now ends with `cp27-cp27m-macosx_10_10_intel.whl`, whereas before this change it was `cp27-cp27m-macosx_10_12_intel.whl`.

anthrotype:
If instead of using the system python you install python from python.org, you will get 10.6 as minimum supported python for the generated wheel.
See how multibuild installs Mac python from python.org

anthrotype:
https://github.com/matthew-brett/multibuild/blob/devel/osx_utils.sh#L156

miguelsousa:
Good to know, but I think 10.10 is far back enough.

miguelsousa:
The wheel that got deployed with the [`2.7.0c5`](https://github.com/adobe-type-tools/afdko/releases/tag/2.7.0c5) tag is for 10.12  🤔
I'm guessing `osx_image: xcode6.4` is also needed in the **OSX-deploy** environment.

Issue #335, opened 2018-04-24
[spot] MacOSX 64-bit build fails to dump any font


blueshade7:
Run MacOSX 64-bit **spot** binary as:
```spot MinionPro-Bold.otf```

Result:
```spot [FATAL]: bad input object size [8]```

readroberts:
I can't reproduce this, either the 2012 MinionPro-Bold.otf version 2.115, or with Minion3-Bold.otf version 1.006. Can you check that the spot you are running was built as 64 bit, and point me to the MinionPro-Bold.otf  that you saw this with?

readroberts:
PS. Are you building with the '64bit-support' branch?

miguelsousa:
Fixed via #342

PR #342, opened 2018-04-26
fix for #335 [spot] MacOSX 64-bit build fails to dump any font

blueshade7:
MacOSX 64-bit build of spot, sfntdiff, sfntedit were not working because long type is 64-bit on this platform when expected to be 32-bit.
Modified numtypes.h in spot and sfntdiff to include public/api/supportpublictypes.h for integer types based on int32_t etc.
sfntedit was using raw "unsigned long". Replaced it with Card32.
Removed "l" from printf formats where 32-bit integers are passed.
Made sure Windows and Linux targets build as well.


cjchapman:
@miguelsousa `Squash and merge` or `Rebase and merge`?

miguelsousa:
squash

PR #344, opened 2018-04-27
64bit support

miguelsousa:


readroberts:
@miguelsousa I am glad to see that this PR can be merged automatically. I was not expecting to merge it into the main branch just yet, as there are two outstanding issues. 1) makeotf has a bug in processing seac operators in the source file, 2) rotateFont has a bug in setting the FontBBox. I currently plan to work on these as soon as I get back to work on Monday afternoon.

miguelsousa:
@readroberts commit 8b6e1d75116b6c2f6f9e1b82358e536cedaae474 is just whitespace change whereas the commit message is *[rotateFont] Fix very old bug in rotating the FontBBox*.

anthrotype:
@miguelsousa the last tag you just pushed “2.7.0rc” is normalised to a PEP440-compatible version string “2.7.0rc0”, however the latter compares older than the preceding tag which was 2.7.0c6 (“c” is equivalent to “rc”). I’d remove the tag, rerelease as 2.7.0rc7

miguelsousa:
👌

PR #355, opened 2018-05-03
[tx] Fix for issues 353 (tx assumed that a CFF2 font is a variable fo…

readroberts:
…nt) and 354 (tx had an array overflow with many hints).

1) Not only did tx assume that in a CFF2 table, the VariationStore is non-zero, it also did not not check in other places.
2) The array allotted for blended arguments was sized only for path operations, and was not large enough for stem arguments.


miguelsousa:
I've changed the base branch to be `develop` instead of `master`.

miguelsousa:
@readroberts what's the status on this PR?

readroberts:
I just forgot about it. Will deal with this when I get back from work on Weds.

readroberts:
Noted. Will get to  Tues-Weds this week.

PR #377, opened 2018-05-15
Reenable tests

miguelsousa:


miguelsousa:
This time AppVeyor had no problems with the same exact `ttxn` tests that made #375 fail. This doesn't make sense to me; I don't know what is causing the tests to fail on some of the times.

I should mention that it's always the **pr** build that fails erratically; I haven't seen the **branch** build fail for unexpected reasons.

@cjchapman unless you have a better solution, I think that going forward we can ignore AppVeyor's **pr** failures as long as the **branch** build is successful. Having to deal with these mysterious failures has cost me a bunch of time.

I'm going to re-enable more tests and see what happens. All bots are green right now.

Issue #373, opened 2018-05-14
[spot] Abort trap with spot -t GSUB=7 <ttf_font>


miguelsousa:
**Abort trap** when running the command `spot -t GSUB=7 <font>` on a TTF file that contains a long glyph name. If the font is OTF all is fine.

miguelsousa:
[Test case](https://github.com/adobe-type-tools/afdko/blob/develop/Tests/spot_test.py#L111-L117)

miguelsousa:
Fixed by #396

Issue #371, opened 2018-05-14
[spot] Segmentation fault with spot -t GPOS=6 <font>


miguelsousa:
**Segmentation fault** when running the command `spot -t GPOS=6 <font>`

miguelsousa:
[Test case](https://github.com/adobe-type-tools/afdko/blob/develop/Tests/spot_test.py#L60-L65)

Issue #244, opened 2018-01-20
[autohint] conflict problem when run simultaneously


kawabata:
Looking at autohint.py
````python
	# temp file names for input and output bez files, and for the fontinfo file.
	tempBaseName = os.path.join(tempfile.gettempdir(), "fdkautohint")
	tempBez = tempBaseName + ".bez"
	tempBezNew = tempBez + ".new"
	tempFI = tempBaseName + ".fi"
````

This seems fixed name temp file is created, causing conflicts when multiple
autohint programs run simultaneously.  It may be better create the temp file name
with proper procedure (e.g. use tempfile) .

miguelsousa:
I totally agree.

typemytype:
I did not fix this while porting to py3, tried to keep the original behaviour 

see https://github.com/adobe-type-tools/afdko/blob/f1c4511eaf6ae99fe784c516f16a3b8ae6a995c9/afdko/Tools/SharedData/FDKScripts/autohint.py#L1063-L1066

miguelsousa:
Confirm that this does not happen in `psautohint`.

PR #380, opened 2018-05-16
Spot gpos fixes

readroberts:
Fixes for issues 373 and 371 in dumping GPSO table with spot.

### DIFF
--- a/afdko/Tools/Programs/spot/source/GDEF.c
+++ b/afdko/Tools/Programs/spot/source/GDEF.c
@@ -253,7 +253,7 @@ static void dumpAttachList(Offset attachListOffset, AttachList * attachList, int
 
 	for (i = 0; i < attachList->GlyphCount; i++)
 		{
-		char name[64];
+		char name[MAX_NAME_LEN];
 		GlyphId glyphId = *da_INDEX(CovList.glyphidlist, i);
 		AttachPoint * attachPoint = &attachList->_AttachPoint[i];
 
@@ -311,7 +311,7 @@ static void dumpLigCaretList(Offset ligCaretListoffset, LigCaretList* ligCaretLi
 	
 	for (i = 0; i < ligCaretList->LigGlyphCount; i++)
 		{
-		char name[64];
+		char name[MAX_NAME_LEN];
 		CaretValueFormat3 * caretValue;
 		GlyphId glyphId = *da_INDEX(CovList.glyphidlist, i);
 		LigGlyph * ligGlyph = &ligCaretList->_LigGlyph[i];
--- a/afdko/Tools/Programs/spot/source/GPOS.c
+++ b/afdko/Tools/Programs/spot/source/GPOS.c
@@ -46,9 +46,10 @@ static IntX GPOSLookupIndex = 0;
 static IntX GPOSLookupCnt = 0;
 static IntX GPOSContextRecursionCnt = 0;
 
-static Byte8 contextPrefix[40]; /* when dumping the context sub rules, this cntains the context string for the rule, if any. */
+static Byte8 contextPrefix[MAX_NAME_LEN]; /* when dumping the context sub rules, this cntains the context string for the rule, if any. */
 
 static FILE * AFMout;
+static char* tempFileName;
 
 static void *readSubtable(Card32 offset, Card16 type);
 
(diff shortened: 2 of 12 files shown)


miguelsousa:
Something is going on with Codacy.

cjchapman:
@readroberts Doesn't Python 2.7 use VS 2010 to compile C code on Windows (i.e. when you build everything through setup.py)?

readroberts:
@msousa @cjchapman. Yes, Travis is using VS 2010 to compile C code on Windows.  I wouldn't accept this PR as is until I update the Windows builds to VS 2017. The alternative is to write some simple code for Windows only ( development time about 15 minutes - would just create a temp file in the current directory, without attempting to deal more than one version of 'spot' running) and then take it out when we get around to updating all the Windows projects. I favor the latter.

miguelsousa:
@readroberts how's updating the project files to VS 2017 going to make a difference given that (AFAIK) VisualStudio hardwires the version of Python it call? Our Python scripts are not ready to go beyond Python 2.7 which (AFAIK) means we're stuck on VS 2010.

readroberts:
@msousa @cchapman Not on our system: we don't use Cython to build extensions. In the AFDKO setup.py, I override setuptools.command.build_py.build_py module with a custom class 'CustomBuild()'. The run() method in this class calls shell scripts that compile the code. The Windows 'BuildAll.cmd' command file references the build system for  specific version of Visual Studio. I don't know how Travis works, but I think I recall Chis telling me that it can be configured to run on whatever Windows configuration you want.

cjchapman:
@readroberts When I did `python setup.py build` on my Windows 10 VM yesterday, it failed because I didn't have VisualStudio 2010 installed, and only built after I installed VisualStudio 2010.

anthrotype:
strictly speaking, if you were to build an extension module for python 2.7 on windows, you would need visual studio 2008, which is ancient. MS provides a package called "Visual C++ for Python 2.7" that includes only the commandline tools from vs2008 and can be used to compile python extension module for 2.7. For compreffor, that was not too old, so I had to resort to using mingw-w64 toolchain (basically GCC on Windows).
Python 3.3-3.4 series uses vs2010, python 3.5-3.6 use vs2015 and the new 3.7 uses vs2017 (if I'm not mistaken).

anthrotype:
(typo -- of course I meant, "that _was_ too old"...)

anthrotype:
> Travis is using VS 2010 to compile C code on Windows

you meant Appveyor. All the available visual studio versions should already be installed there, see
https://www.appveyor.com/docs/build-environment

miguelsousa:
I believe the next step is for @readroberts to update **spot**'s Visual Studio files.

readroberts:
@miguelsousa @cjchapman After some experimentation, I see that we cannot  build a mix of VS 2010 and VS 2017 Windows projects.  The default image used by Appveyor is the VS 2015 image, which has the VS 2010 .Net build tools that is used by our Windows projects. However, the VS 2017 image does not have this. If I want to update any of our windows projects to VS 2017, I need to update all of them. This is pretty easy, and I expect that I could do it in one day. Going ahead with this is my preference, but I could also revert to plan B: for the moment, continue to use VS 2010, but use a simple and reasonably reliable method for generating a temp file. With plan B, we defer updating the Windows projects until later.

miguelsousa:
> If I want to update any of our windows projects to VS 2017, I need to update all of them. This is pretty easy, and I expect that I could do it in one day. Going ahead with this is my preference

Updating to VS 2017 would be my preference too, but your time estimates are **always** over-optimistic, so it's difficult for me to weigh what the best option is.

readroberts:
The simple approach to generating a temp file is to create a temp file with hard-coded name like "spot.tmp' in the current directory, without attempting to deal more than one version of 'spot' running, and then reverting this when we have updated to VS 2017.  The work involved in updating to to VS 2017 is to 1) open each of 12 Visual Studio solution  files in VS 2017, and compile it, then 2) edit each of the 12 BuildAll.cmd files to reference the VS 2017 tools.

miguelsousa:
The temp file hack doesn't seem like a good option to me; I don't want us to end up with another issue like #244.

Here's my suggestion of what should be done:
1. Create a new PR with the changes in 3c85e25d744875bfc96f6dc3ce58de500405affc; just start a new branch from `develop` and cherry-pick this commit. That PR will address #373.
2. Close this current PR.
3. Open a new issue about converting the Windows builds to VS2017.
4. When you're ready, create a new PR containing the changes to address point 3 and issue #371.

anthrotype:
Requiring such a recent version of Microsoft's toolchain just to get a working `tmpfile` implementation seems a bit too much to me, especially for a code base that is basically all ANSI C.
I'm no C programmer, but I bet somebody else had the same issues and worked around it somehow. And you could probably find alternative ways to communicate between processes than temporary files, like using stdout/stdin.
I'm all for the new and shiny,I'm only concerned that you may be tempted to introduce dependencies on features only available on such recent compiler, and then it could become harder to compile AFDKO C modules in situations where the compiler version is constrained for some reasons (e.g. when making CPython extensions).
Anyway, just my 2c. You do what you think is best :)

PR #385, opened 2018-05-21
[ttxn] Fix shell calls to 'tx'

miguelsousa:


miguelsousa:
@cjchapman the `ttxn` changes had no effect on the AppVeyor failures, but I think this fix is worth merging.

PR #396, opened 2018-05-24
[spot] Fix a crash due to buffer overrun errors from long glyph names.

readroberts:
With the 64-bit update, the function global.c::getGlyphName() can return a glyph name of up to 128+7 characters long. The returned string was, in many places, copied into a string buffer that was much shorter. Fixes issue 373.

### DIFF
--- a/afdko/Tools/Programs/spot/source/global.h
+++ b/afdko/Tools/Programs/spot/source/global.h
@@ -113,6 +113,8 @@ extern IntX parseIdList(Byte8 *list, IdList *ids);
 extern IntX getNGlyphs(Card16 *nGlyphs, Card32 client);
 extern void quit(IntN status);
 extern void initGlyphNames(void);
+#define NAME_LEN 128 /*  base glyph name length returned by getGlyphName */
+#define MAX_NAME_LEN  NAME_LEN+7 /* max glyph name langth returned by getGlyphName */
 extern Byte8 *getGlyphName(GlyphId glyphId, IntX forProofing);
 extern void dumpAllGlyphNames(IntN docrlfs);
 extern void getMetrics(GlyphId glyphId, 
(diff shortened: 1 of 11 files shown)


afdko/Tools/Programs/spot/source/global.h
```diff
@@ -113,6 +113,8 @@ extern IntX parseIdList(Byte8 *list, IdList *ids);
 extern IntX getNGlyphs(Card16 *nGlyphs, Card32 client);
 extern void quit(IntN status);
 extern void initGlyphNames(void);
+#define NAME_LEN 128 /*  base glyph name length returned by getGlyphName */
+#define MAX_NAME_LEN  NAME_LEN+7 /* max glyph name langth returned by getGlyphName */
```
miguelsousa: Typo langth

afdko/Tools/Programs/spot/source/opt.c
```diff
@@ -285,7 +285,7 @@ int opt_##n(int argc, char *argv[], int argi, opt_Option *opt) \
 	else \
 		{ \
 		t value; \
-		char s[64]; \
+		char s[128]; \
```
miguelsousa: Should `NAME_LEN` or `MAX_NAME_LEN` be used here?

readroberts: Didn't mean to change this. I will revert it. There is no dependency on  glyph name length here.  This code is for parsing the  arguments. 's' holds values for arguments, like "GPOS=6". There is only one potentially long string value. This is the list of glyph ID's used with the  command "spot -t CFF=6 -g <list of glyph ID's>. This  cannot take glyph names, or I would have had to allocate memory dynamically. The logic keeps the buffer from being overrun by truncating the input string, if it is too long.

PR #407, opened 2018-06-05
Fixes for Tests under Windows.

readroberts:
In differ.py, when 'differ' walks a dir tree to make a list of files in a ufo file, Windows and Mac list the files in a different order. The lists are compared, and the diff fails before the files are even examined.

in tx_test.py, when pdf files are compared, an additional line in the pdf file needs to be ignored. This didn't cause a problem before, but as of June 1st, the date string changed enough to make this extra line change.


Tests/differ.py
```diff
@@ -163,6 +163,7 @@ def _get_all_file_paths_in_dir_tree(start_path):
 
         # Make the paths relative
         all_paths = [path.replace(start_path, '') for path in all_paths]
+        all_paths.sort()
```
miguelsousa: I don't think this is necessary because the paths are not going to be compared between Win and Mac

readroberts: Sorry, I made a mistake in the comment - the difference is between the reference directory tree and the new temporary director tree when the test is run under Windows, not Mac and Windows.

miguelsousa: OK, then use `sorted()` instead.

Tests/differ.py
```diff
@@ -196,6 +197,9 @@ def _paths_are_same_kind(path1, path2):
 
 
 def _validate_path(path_str):
+    if path_str is None:
+        raise argparse.ArgumentTypeError(
+            "{} is not a valid path.".format(path_str))
```
miguelsousa: I don't think this validation is needed because this method will only be called whent the command line argument is provided; if it's not provided things will fail earlier

readroberts: @miguelsousa   I put the code in because there is a path where this not caught. In runnner.run_test(), if either of the subprocess calls encounters an exception, it is caught and the function returns None as the output path. Most of the test functions receive this as 'actual_path', and then call differ().  In differ.main(), the get_options() call converts the "None" to the current directory. The result is a mis-leading error message that the path types differ, as the expected_path is a file, and the actual_path is a directory.

miguelsousa: OK, in that case please add a test case for this situation.

miguelsousa:
@readroberts don't forget to bump the versions of `differ.py` and `runner.py`

miguelsousa:
@readroberts all the commits to `win-tests-1` branch built fine on AppVeyor, but when the branch got merged into `develop` it failed. All the failures seem to be related to **tx**.
https://ci.appveyor.com/project/adobe-type-tools/afdko/history

readroberts:
@msousa @cchapman I still can't reproduce this on my local Windows system. I see the the errors are only with the '-dump' option of tx. This is what is causing the failures in the ttxn command and the fontplot commands as well. I might have suspected a problem with writing to stdout, but in most of the errors, an output file is supplied as an explicit option to tx. When I get back to work Mon afternoon, I will focus on trying to generate a reproducible crash case.

Issue #408, opened 2018-06-07
Installing may downgrade dependencies


moyogo:
In `setup.py`, `install_requires` is filled with the content of [`requirements.txt`]( afdko/requirements.txt).
https://github.com/adobe-type-tools/afdko/blob/develop/setup.py#L205
https://github.com/adobe-type-tools/afdko/blob/develop/setup.py#L145-L147

These requirements dependency versions are pinned with version matching, which means some packages may be downgraded to those versions if more recent versions happen to be installed.

Could these be changed to inclusive ordered comparison `>=` or compatible release `~=` instead of version matching `==`?

cjchapman:
@moyogo I'm OK with either approach, but I'm not sure what @miguelsousa's thoughts are (as he's on leave right now), so for now I've submitted a PR to check for compatible releases (`~=`). Once I hear from him, if he's OK with ordered comparison (`>=`), I'll make that change.

In related news, I sent you a collaborator invite. Sorry for not doing this sooner, I had assumed you were already on the list until I tried to add you as a reviewer to the PR.

readroberts:
I think this is a good idea. Definitely for modules like setuptools. I am am ambivalent about the modules related to font tools, as these move quickly, and at any moment may break the afdko. However, I overall favor taking the latest, and taking the work fo keeping up.

anthrotype:
https://packaging.python.org/discussions/install-requires-vs-requirements/

> It is not considered best practice to use install_requires to pin dependencies to specific versions, or to specify sub-dependencies (i.e. dependencies of your dependencies). This is overly-restrictive, and prevents the user from gaining the benefit of dependency upgrades.

For more on “Abstract” vs “Concrete” requirements, see https://caremad.io/2013/07/setup-vs-requirement/.

moyogo:
@cjchapman Thanks!

I think pyup-bot will restore the `==` in its next PR. Maybe these need to be changed when `requirements.txt` is read to fill `install_requires`.

anthrotype:
no, I think pyup bot leaves open ended requirements as they are, unpinned. Only updates those that are `==`

miguelsousa:
I'm fine with unpinning as long as that doesn't block pyup-bot from updating the version numbers.

pyup-bot runs every Monday, we've been issuing pre-release builds regularly, and we're getting closer to having more frequent official releases; so I don't expect the versions of the dependencies to lag that much.

[...]

PR #409, opened 2018-06-07
fix for #408: change requirements files to check for compatible versions

cjchapman:
(instead of exact version match)


cjchapman:
The build errors are due to test bugs we're trying to fix in https://github.com/adobe-type-tools/afdko/pull/407

Issue #18, opened 2014-11-04
debug and release?


holomorph:
There is a ton of stuff (Makefiles) duplicated where the only difference is the word "debug" or "release" and a `-g` in CFLAGS.  The shell wrapping (`FOO/build/linux/gcc/BuildAll.sh`) unconditionally builds both. Why?  It's simple enough to just build with debug symbols, then strip them out if you don't want them.

HinTak:
Putting a -g then strip, is not the same as doing without -g - bugs in compiler and and strip aside. There are bugs in compilers from time to time where only the result with -g or without it works, and there are the occasional bugs with strip also, where applying strip actually cripples your binary. Those are side issues.

Anyway, the difference, in principle, is that if you do -g, you think there is a possibility the outcome might be broken and you intend to look into it. -g disables some optimizations which might make debugging impossible, as a general rule. Major inlin'ing, code re-arrangements, etc. So as a general rule, the resulting binary from -g then strip will still be slower than without -g. (assuming your compiler is working correctly in both cases).

That's why, depends on circumstances, you want to do both, or either.

HinTak:
Alternatively, in windows' land, it is routine to ship only "release" binaries, for exact this reason: to make running and poking into the binary with a debugger difficult or impossible. On the typical linux box, debug info are stripped, but stored externally, so obviously it is still possible.

HinTak:
In fact -g then strip is perhaps worse than either plain -g or without -g: you don't get a fully optimized binary, and you cannot debug with it either.

I'd say, if you have to choose one, make up your mind which is more important - fully speed or debugg'ability.

miguelsousa:
Fixed via #414

PR #414, opened 2018-06-12
Build Script Update.  Fix all the BuildAll.cmd and *.sh files

readroberts:
... to build only the release projects if no target is given.

They will also run the right build for the targets 'release', 'debug', and 'clean', and do nothing for any other target.


readroberts:
@msousa @cchapman  The build scripts worked - it is the usual 'tx -dump' tests that fail.

miguelsousa:
Related to #18

PR #415, opened 2018-06-13
Remove SunOS relics

### LINKED ISSUE #399: Remove Sun4 code
Doing a search with `sun4` and `PLAT_SUN4` reveals many hits.

miguelsousa:
Fixes #399


afdko/Tools/Programs/public/lib/api/supportenvironment.h
```diff
@@ -285,7 +283,7 @@ that changes to environment.h be upward-compatible, as described above.
  */
 
 #ifndef PROTOTYPES
-#define PROTOTYPES	(OS!=os_sun)
+#define PROTOTYPES	(OS)
```
cjchapman: That's not how this is meant to be used -- see the comment block above this line for more information about it, but it doesn't seem to be used at all, so I think you should remove this and the comment block above it.

miguelsousa:
@cjchapman another pass done

afdko/Tools/Programs/public/lib/api/atm.h
```diff
@@ -62,16 +62,7 @@ buildch and other ATM modules (font parser).
...
 
-/*
- * Some of the PSENVIRONMENT configurations won't compile with ENABLE_ASSERT,
- * so we only turn it on for isp_sparc.  
- * This is enough to find machine-independent PSENVIRONMENT problems.
- */
-#if STAGE==DEVELOP && ((defined PSENVIRONMENT && !PSENVIRONMENT) || ISP==isp_sparc)
-#define ENABLE_ASSERT 1
-#else
 #define ENABLE_ASSERT 0
```
cjchapman: I think we want to enable asserts, and this should probably be:
```c
#if STAGE==DEVELOP && (defined PSENVIRONMENT && !PSENVIRONMENT)
#define ENABLE_ASSERT 1
#else
#define ENABLE_ASSERT 0
#endif
```

miguelsousa:
@cjchapman done. OK to merge?

PR #424, opened 2018-06-15
AppVeyor experiment: added a check for the NULLs I added in tx.c

cjchapman:
this is an experimental PR, please do not merge


cjchapman:
We're getting the same version of `tx.c` in both the branch and PR AppVeyor builds, so that's not the issue.

PR #425, opened 2018-06-15
fixes for modules used by tx

cjchapman:
@miguelsousa (et al.), I'll be making a series of changes in `uforead.c`, cleaning up various issues in the code, in the hopes of fixing `tx`. I'm doing this one step at a time to check the effect each change has on AppVeyor (and Travis). I'll let you know when I'm done.

UPDATE: I'm broadening the scope to include other modules used by `tx`.


miguelsousa:
@cjchapman I'm going to add a test case for the trademark string handling

miguelsousa:
@cjchapman I think this is ready to merge

PR #434, opened 2018-06-19
[tests] Fix absolute imports

miguelsousa:


miguelsousa:
Trying to address these notices
```
Coverage.py warning: Module Tests/runner.py was never imported. (module-not-imported)
Coverage.py warning: Module Tests/differ.py was never imported. (module-not-imported)
```
that show up on Travis after I [added **runner.py** and **differ.py** to the config file](56befefb4b3f6daa26d670d05e3c95df4ec32dd5).

The warnings are still showing up after these changes 🤔

Issue #444, opened 2018-06-25
[tx] Crash with CJK variable font


miguelsousa:
Command: `tx -0 CJK-VarTest.otf`
Test file: [CJK-VarTest.otf.zip](https://github.com/adobe-type-tools/afdko/files/2134955/CJK-VarTest.otf.zip)

Result:
```
tx: --- CJK-VarTest.otf
tx: (cfr) bad INDEX offset array
Segmentation fault: 11
```

miguelsousa:
TTX (v3.28.0) and FontView (v0.3.0) have no problem handling this font.

cjchapman:
The charstring for the glyph for U+8619 is too long according to the [CFF2 CharString Implementation Limits](https://docs.microsoft.com/en-us/typography/opentype/spec/cff2charstr#appendix-b-cff2-charstring-implementation-limits), which gives the "CharString length" limit as 65,535 bytes and this glyph's charstring length is 71,057 bytes.

However, Sairus Patel pointed out that this is an implementation limit, not a spec. limit per se. The `tx` program is currently treating it as if it were a spec. limit, which we plan to change.

As an aside, @blueshade7 pointed out that there are many integer values in the glyph that were stored as floating point numbers, which is very inefficient. When I converted them all to be stored as integers, the length of the glyph dropped below the 65,535 byte limit.

So, to summarize, the font is not making efficient use of glyph charstring bytes, but it is still a valid font.

miguelsousa:
@cjchapman the hard limit in tx should only be lifted for CFF2 fonts, right?

cjchapman:
@miguelsousa I think that limit may also be an "implementation limit" rather than a "spec limit" for CFF, not just CFF2, but let me check with Sairus.

blueshade7:
I do not recall of any implementation that inherently had this 65535 limit, but by enforcing it, AFDKO and any products using it became THE implementation having this limit. A Japanese proverb puts it: the mummy hunter himself became a mummy.

I believe we should change fatal to message as a warning.
Another side of the problem is that tx shouldn’t crash when fatal is called. It should gracefully exit.

miguelsousa:
charstring limit addressed by #446

sairuspatel:
Note that CoolType apps will currently run up against the charstring length limit of 65535. So it’s appropriate that tx emit a warning.

cjchapman:
@sairuspatel As of PR #446, `tx` will give a message like:
[...]

PR #446, opened 2018-06-27
fix for issue #444: make CharString length check a warning

cjchapman:
(rather than a fatal error)


miguelsousa:
@cjchapman I think I know why Travis didn't kick in. Looking into it now.

Tests/tx_test.py
```diff
@@ -174,3 +174,11 @@ def test_remove_hints_bug180():
...
     expected_path = _get_expected_path('cid_nohints.ps')
     assert differ([expected_path, actual_path, '-m', 'bin'])
+
+
+def test_long_charstring_bug444():
+    # read a CFF2 VF with a charstring longer that 65535, check output
+    actual_path = runner(CMD + ['-o', '0', '-f', 'CJK-VarTest.otf'])
+    expected_path = _get_expected_path('CJK-VarTest.txt')
+    skip = ['-s', '## Filename']
+    assert differ([expected_path, actual_path] + skip)
```
miguelsousa: you don't need to define `skip`

cjchapman: I think I do. If I comment it out, I get:
```
    def test_long_charstring_bug444():
        # read a CFF2 VF with a charstring longer that 65535, check output
        actual_path = runner(CMD + ['-o', '0', '-f', 'CJK-VarTest.otf'])
        expected_path = _get_expected_path('CJK-VarTest.txt')
        # skip = ['-s', '## Filename']
>       assert differ([expected_path, actual_path] + skip)
E       NameError: global name 'skip' is not defined
```

miguelsousa: sorry what I meant was, just do
```python
assert differ([expected_path, actual_path, '-s', '## Filename'])
```

Issue #457, opened 2018-07-02
makeotf option typo (afdko 2.7.2)


LIXiangChen:
`makeotf -h` said:

> `-shw/-nshw` Do/Do Not suppress warnings about unhinted glyphs. This is useful when processing unhinted fonts.

However, `-shw` is actually DO NOT show warning, and `-nshw` will thrown `makeotfexe [FATAL] unrecognized option (None)`.

Issue #264, opened 2018-02-06
Move 'Adobe Cmaps' and 'CID charsets' to separate repositories


miguelsousa:
Move **Adobe Cmaps** and **CID charsets** to separate repositories. These assets would then be loaded as git modules.

Making these separate repositories will allow other tools, such as [fontmake](https://github.com/googlei18n/fontmake), to build CID-keyed fonts. It will also enable versioning.

@kenlunde can you take care of this?

kenlunde:
The content of **Adobe Cmaps** is kept up-to-date in the [cmap-resources](https://github.com/adobe-type-tools/cmap-resources/) repository, and the following two patterns can be used to load only the ones that are necessary for CID-keyed OpenType/CFF font development:

```
*/CMap/*UTF32-H
*/CMap/*[35BC]p[cv]*-H
```

The above resolves to:

```
Adobe-CNS1-7/CMap/B5pc-H
Adobe-CNS1-7/CMap/UniCNS-UTF32-H
Adobe-GB1-5/CMap/GBpc-EUC-H
Adobe-GB1-5/CMap/UniGB-UTF32-H
Adobe-Japan1-6/CMap/83pv-RKSJ-H
Adobe-Japan1-6/CMap/UniJIS-UTF32-H
Adobe-Japan1-6/CMap/UniJIS2004-UTF32-H
Adobe-Japan1-6/CMap/UniJISX0213-UTF32-H
Adobe-Japan1-6/CMap/UniJISX02132004-UTF32-H
Adobe-Korea1-2/CMap/KSCpc-EUC-H
Adobe-Korea1-2/CMap/UniKS-UTF32-H
```

The content of **CID charsets** is a bit complex in that almost all of the files are not actually used by public AFDKO tools, but rather by a non-public tool that is used to compile CIDFont resources from a series of Type 1 (PFA) row fonts. Let me ponder this.

miguelsousa:
@kenlunde any decision on the **CID charsets** files? I'd like to resolve this issue.

kenlunde:
@miguelsousa The **CID charsets** folder and its contents can be removed with extreme prejudice, because none of the public AFDKO tools reference or require them. They are for internal (Adobe) use.

Issue #222, opened 2017-12-19
Don’t limit folder paths to ASCII


frankrolf:
Currently, makeOTF will fail if a containing folder contains a non-ASCII character.
This comes up often in my workshops, and also in real life:
https://github.com/adobe-fonts/source-han-sans/issues/189

anthrotype:
filesystem paths are byte strings encoded using sys.getfilesystemencoding() on python 2. On python 3 they are (unicode) strings (internally decoded using sys.getfilesystemencoding() on posix platforms, or passed as such from Windows “wide” console API).

The filesystem encoding is platform dependent. On most unix systems (including macs) is UTF-8, on Windows is some code page depending on the locale.

miguelsousa:
@anthrotype how does **fonttools** deals with this kind of issue? I can't find any call to `sys.getfilesystemencoding()` in it.

anthrotype:
the filesystem encoding is only needed if you want to decode the byte strings that you get from (python2) sys.argv to a unicode string.
If you get a path from sys.argv and use it to, say, `open()` the file, it should work fine.

We need to check if somewhere in makeotf that byte string path is being coerced to a unicode string. This happens for example when you concatenate a bytes string and a unicode string, in python 3 it's an error, whereas on python 2 the bytes string gets upgraded to unicode using the sys.getdefaultencoding() which is normally "ascii".

Would be good to also know what operating system the issue noted by @frankrolf happened.

Many of these unicode issues are fixed by just upgrading to python3, but I undestand this option is available for everyone.

anthrotype:
I think I found where the coercion from bytes to unicode is happening.
 
https://github.com/adobe-type-tools/afdko/blob/71c8d99f088fcf8b1159584d53937ade650534ea/FDK/Tools/SharedData/FDKScripts/MakeOTF.py#L781

in order to get the current working directory, you're using `os.getcwdu` which returns a unicode string, instead of `os.getcwd` that returns a bytes string:
https://docs.python.org/2/library/os.html#os.getcwdu

Then this unicode `currentDir` is used later, for exampe inside `writeOptionsFile`, and is concatenated with other _bytes_ strings, which may contain non-ascii characters. When this happens, the bytes string is implicitly decoded to unicode using the default python encoding (not to be confused with the filesystem encoding) which is "ascii". In general, it's preferable to avoid this implicit coercion and instead only concatenate unicode to unicode or bytes to bytes, and do the decoding/encoding explicitly.

[...]

PR #458, opened 2018-07-03
Makeotf fixes

### LINKED ISSUE #148: makeotf: "[Warning] the feature file does not contain a 'vert' feature"
If the `vert` feature is written with the exception of "main feature file" by calling include command, makeotf will prompt "[Warning] the feature file does not contain a 'vert' feature"

miguelsousa:
Fixes #148, #222, #264, #457 and #459


.travis.yml
```diff
@@ -3,7 +3,7 @@ matrix:
   include:
     - os: linux
       env: NAME=Linux
-      if: type = pull_request
+      if: NOT branch =~ /^\d+\.\d+\.\d+.*$/
```
cjchapman: Please add detailed comments here and for each of the changes below so that future generations may learn more quickly than we did how and why to do this kind of thing.

cjchapman: (Also, there are lots of minor issues that I'd like for us to clean up in MakeOTF.py, but nothing specific to the changes.)

miguelsousa: done

Issue #456, opened 2018-07-01
[makeotfexe] Generates OTFs with ridiculously large glyph advance widths


miguelsousa:
When a glyph in a source font has a negative advance width (`-x`), **makeotfexe** will set the advance width of that glyph in the resulting OTF to `65536 - x`.

```
makeotfexe -f negative.pfa
```
[negative.pfa.zip](https://github.com/adobe-type-tools/afdko/files/2153131/negative.pfa.zip)

readroberts:
It is correct that makeotfexe writes a negative advance width value as a signed short (well, it writes it as an unsigned short, but does so after simply casting the the width value to an unsigned short, so you get the same two bytes as if you had written a signed short).  65536 - x is what you get if you read a negative signed short value as an unsigned short.  I am a little concerned about making this illegal, as developers have used negative widths in some cases - I vaguely remember a discussion about Semitic fonts using negative widths; it was certainly used to get around bugs in GPOS in the early days of OpenType. If the consensus is that it is time to disallow negative values, would you 
 prefer to have negative advance widths set to 0, with a warning, or a fatal error?

miguelsousa:
The negative advance width should be set to 0, and a warning should be issued.

PR #460, opened 2018-07-05
[makeotfexe] Fix issue #456: negative widths are written as large numbers.

readroberts:
makeotfexe wrote negative values into an unsigned short field, resulting in a value of (65536 - value). makeotfexe now sets negative values to 0, and issues a warning.


miguelsousa:
@readroberts thanks, I'll takeover this branch. #458 needs to be merged first.

PR #464, opened 2018-07-10
Update ufoLib to 2.2.4

miguelsousa:


anthrotype:
I don't know if you also wish to pin the lxml version, now that ufoLib >= 2.2 requires that one.
I just mention this for completeness. If you don't add it, it will be installed anyway (whatever min version ufoLib itself requires, and is available at the moment of installing ufoLib).

miguelsousa:
@anthrotype the XML header is different. `<?xml version="1.0" encoding="UTF-8"?>` versus `<?xml version='1.0' encoding='utf-8'?>`. This is a problem. Not just for the tests but also for the thousands of  UFOs under version control.

anthrotype:
i'm sorry. That's the header that lxml (via libxml2) produces, we can't change that. This is a consequence ufoLib shifting from a custom-made xmlwriter to lxml. Feel free to file an issue over at ufoLib. I think this is not such a big issue. Adding hacks to workaround it would only complicate things further. The two xml declaration are identical as far as XML specification.

anthrotype:
this is ugly but if you do really care, I'd accept a PR for this:
https://stackoverflow.com/a/47132033

khaledhosny:
May be doing XML-aware diff would help, something like
https://pypi.org/project/xmldiff/ (I haven't tried it).

>

miguelsousa:
How severe this change is will depend on who you ask. I've opened https://github.com/unified-font-object/ufoLib/issues/158 not so much to request a rollback to the previous XML header but mostly to document the change and to have something we can reference and discuss.

PR #467, opened 2018-07-11
Issue354 fixes

readroberts:
Add test for issue 354 " Fails to dump a CFF2 VF font with many hints". Also fixed bug in writing CFF2

In writing test for issue 354, discovered that tx was writing the CFF2 var store incorrectly when writing a CFF2: it was adding the var store offset twice during offset calculations.


Tests/tx_test.py
```diff
@@ -196,3 +196,13 @@ def test_long_charstring_bug444():
     actual_path = runner(CMD + ['-o', '0', '-f', 'CJK-VarTest.otf'])
     expected_path = _get_expected_path('CJK-VarTest.txt')
     assert differ([expected_path, actual_path, '-s', '## Filename'])
+
+def test_many_hints_string_bug354():
+    # The glyph T@gid002 has 33 hstem hints. This tests a bug where
+    # tx defined an array of only 6 operants.
+    # This is encountered only when wrinting to a VF CFF2.
+    actual_path = runner(CMD + ['-o', 'cff2', '-f', 'cff2_vf.otf'])
+    actual_path = runner(CMD + ['-f', actual_path, '-o', 'dcf'])
```
miguelsousa: I think you need to use `-a` option in this second call to `runner`

miguelsousa:
> In writing test for issue 354, discovered that [...]

This is why I think adding tests is always a good thing.

Tests/tx_test.py
```diff
@@ -196,3 +196,13 @@ def test_long_charstring_bug444():
...
+    expected_path = _get_expected_path('cff2_vf.dcf.txt')
```
miguelsousa: you did not include this file (cff2_vf.dcf.txt) in the commit

Tests/tx_test.py
miguelsousa: Don't use `actual_path` twice, because in case of failure it may be less clear which one is

afdko/Tools/Programs/public/lib/source/cffwrite/cffwrite.c
```diff
@@ -733,10 +733,9 @@ static Offset calcFontOffsets(controlCtx h, cff_Font *font, Offset offset) {
     font->offset.FDSelect =
     cfwFdselectGetOffset(g, font->iObject.FDSelect, h->offset.FDSelect);
 
-    if (font->size.VarStore > 0) {
-        font->offset.VarStore = offset;
-        offset += font->size.VarStore;
-    }
+	if (font->size.VarStore > 0) {
+		font->offset.VarStore = h->offset.varStore;
+	}
```
miguelsousa: fix indents

afdko/Tools/Programs/public/lib/source/cffwrite/cffwrite.c
```diff
@@ -965,14 +964,14 @@ static int calcSetOffsets(controlCtx h) {
...
-            offset += h->size.FDSelect;
-        }
+		if (h->size.varStore > 0) {
+			h->offset.varStore  = offset;
+			offset += h->size.varStore;
+		}
+		if (h->size.FDSelect > 0) {
+			h->offset.FDSelect  = offset;
+			offset += h->size.FDSelect;
+		}
```
miguelsousa: more indent fixes needed

PR #483, opened 2018-07-15
ttxn fixes

miguelsousa:


miguelsousa:
Couldn't get NotoNastaliqUrdu's GSUB dump to work — it seems to do into an infinite loop — otherwise I think it should be ready to merge.

miguelsousa:
@cjchapman I've made lots of fixes and improvements. In the process, learned that what the infinite loop was instead a very inefficient and possibly wrong iteration. I disabled the one causing the major problem as well as other similar iterations, since they're not producing the correct output anyway. Added warnings to the lookups known to have incomplete support.

PR #486, opened 2018-07-16
Update ufoLib to 2.3.0

miguelsousa:


miguelsousa:
The tests won't pass until `mutatorMath` gets an update. See https://github.com/unified-font-object/ufoLib/pull/160#issuecomment-405066945

miguelsousa:
@anthrotype are the data/string differences supposed to happen if ufoLib version 2.3.0 is installed **without** the `lxml` option? I was under the impression that you kept the previous behavior by default.

anthrotype:
I just said that ufoLib now behaves the same way both with or without lxml and in python 2 or python 3. Any of these bytes-that-should-be-unicode should get fixed.

anthrotype:
Of course they only affect python 2.7, where this confusion can happen

miguelsousa:
thanks. Can you confirm my assessment that mutatorMath has bytes-that-should-be-unicode things to fix?

anthrotype:
> Can you confirm my assessment that mutatorMath has bytes-that-should-be-unicode things to fix?

I had underestimated this issue.. Requiring all clients of ufoLib to consistently use `unicode` instead of `bytes` (==`str` in python2) anywhere a plist string can occur is too demanding for old python2 code...

I need to revise my approach in the new `ufoLib.plistlib`. Instead of forcing python2 to behave like python3 (where the text / binary data is clear), I need to implement the `use_builtin_types` option in plistlib.load/dump functions, and add support for the deprecated `plistlib.Data` wrapper class for bytes data.

Python 2 is my nemesis.

miguelsousa:
Thanks for looking into this Cosimo. There will be a time when we don't have to think about Python 2 anymore.

cjchapman:
@miguelsousa @anthrotype I think we should throw a party when that time comes.

benkiel:
> Python 2 is my nemesis.

And we can get t-shirts with this!

miguelsousa:
Or: Python 2 ~~is~~ was my nemesis

benkiel:
> Or: Python 2 ~~is~~ was my nemesis

To state the obvious, the **~~is~~** is crossed out by a snake.

khaledhosny:
Now I got an idea for my next geeky t-shirt, the backtrace of Python 2
encoding exception.

>

Issue #356, opened 2018-05-03
[tx] Fails to convert to CFF a CFF2 variable font with some non-varying glyphs


readroberts:
A glyph which is non-varying in a variable font will be referenced by a VariationStore data item subtable which has a region count of 0. The VF support code assumed that this was an error. I need to correct the title: the conversion to CFF does not fail, but it incorrectly emits an alarming warning "(cfr) out of range region index found in item variation store subtable" for each non-varying glyph.

miguelsousa:
I'll add an option to `runner` to enable saving stderr to a file.

miguelsousa:
Fixed via #476

PR #487, opened 2018-07-16
[tx] fixes for issue 356: a false warning when a CFF2 VF font has some non-varying glyphs

readroberts:
A glyph which is non-varying in a variable font may be referenced by a
VariationStore data item subtable which has a region count of 0. The VF
support code assumed that this was an error, and issued a false warning.


miguelsousa:
I have no idea why this PR was created. It looks like a mashup of issue #356 and its fix (pull #476).

PR #504, opened 2018-07-20
[tx] tx fails when converting a non-VF FF2 OTF font to cff2. Adding test.

readroberts:
Add test for issue 353: tx fails when converting a non-VF FF2 OTF font to cff2.


Tests/tx_test.py
```diff
@@ -176,6 +176,15 @@ def test_varread_pr355():
     assert differ([expected_path, actual_path])
 
 
+def test_cff2_no_vf_bug353():
+    # read CFF2 WITHOUT VF info, write a CFF2 out. 'regular_CFF2.otf'
+    # is derived by taking the regular.otf file from the sfntdiff
+    # 'input_data' directory, and converting the CFF table to CFF2.
```
miguelsousa: How was the conversion from CFF to CFF2 step done? I'm mostly curious.

PR #506, opened 2018-07-22
Update ufoLib to 2.3.1

miguelsousa:


Tests/checkoutlinesufo_data/expected_output/ufo2-dflt-layer.ufo/fontinfo.plist
```diff
@@ -1,162 +1,162 @@
...
-	<integer>1000</integer>
-	<key>versionMajor</key>
-	<integer>1</integer>
-	<key>versionMinor</key>
-	<integer>0</integer>
-	<key>xHeight</key>
-	<integer>474</integer>
-</dict>
+  <dict>
+    <key>ascender</key>
```
anthrotype: @miguelsousa I'm thinking of stripping those leading two spaces and indent the top-level element (usually `<dict>`) the same indentation level as the root `<plist>` element, just like the standard plistlib does...
it seems like a waste of characters.. On the other hand, it may introduce some extra complexity and might slow down dumping a bit.

miguelsousa: They don't bother me, but if you're going to change it, please do it soon.

anthrotype: let's leave them like that then

PR #508, opened 2018-07-23
reorganize directory structure and package structure

cjchapman:
The C source tree will no longer be included in the wheel (`*.whl`) files, and now you can do:
```python
from afdko import <module>
```
instead of:
```python
from afdko.Tools.SharedData.FDKScripts import <module>
```


anthrotype:
much better :+1:

PR #519, opened 2018-07-27
use explicit paths for ignoring compiled OSX & linux programs

cjchapman:
(otherwise changes in matching folders don't get picked up, e.g. in the `tx` folder)

I realized we had a problem after a comment that @anthrotype made in this thread: https://github.com/typemytype/booleanOperations/issues/40#issuecomment-408395042)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,16 +18,8 @@ htmlcov
 .pytest_cache
 
 # ignore the OSX and linux compiled programs
-autohintexe
-detype1
-makeotfexe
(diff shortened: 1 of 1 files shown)


cjchapman:
That does not work. I think because of the `this_folder_intentionally_left_empty` file.

miguelsousa:
This should do it
```
c/build_all/*
!c/build_all/this_folder_intentionally_left_empty
```

Issue #347, opened 2018-04-30
[type1/detype1] Add -h option


miguelsousa:
Add basic `-h` option. Make sure it exists with 0.

The **runner.py** script runs `<tool> -h` and checks the exit code to test if a given tool is available.

miguelsousa:
Fixed via #536 & #537

PR #537, opened 2018-08-08
[type1] Implement '-h' option. Fix exit code for known options

miguelsousa:
Completes fix for #347

### DIFF
--- a/c/type1/source/type1.c
+++ b/c/type1/source/type1.c
@@ -428,8 +428,7 @@ for(;;){
  */
 
 static void usage(void) {
-	fprintf(stderr, "usage: type1 [text [font]]\n");
-	exit(1);
+	printf("usage: type1 [text [font]]\n");
 }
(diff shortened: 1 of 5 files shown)


c/type1/source/type1.c
```diff
@@ -526,7 +530,9 @@ int main(int argc, char *argv[]) {
                 type1(fp1,fp2);
 				fclose(fp1);
 				fclose(fp2);
-	} else
-		usage();
-	return 0;
+    } else
```
cjchapman: Codacy sort of hints at the problem here, which is that you need to add some braces after your `else` and between the `return` statements, i.e.:
```c
    } else {
        usage();
        return 1;
    }
    return 0;
}
```

miguelsousa: 👍 in that case I need to add braces to `detype1.c` as well.

Issue #196, opened 2017-10-03
[makeotfexe] generates incorrect secondary lookups with useMarkFilteringSet flag


NorbertLindenberg:
For the attached font and feature file, makeotf 2.0.94 generates a secondary lookup that has the useMarkFilteringSet bit set, but combines substitutions that use incompatible mark filtering sets. As a result, renderers that pay attention to this bit do not render the font as intended.

The feature file specifies three contextual lookups as part of its pres feature:
– one looking at above base marks while replacing a pre-base mark,
– one looking at below-base marks while replacing a pre-base mark,
– one looking at below-base marks and replacing one of them.

Such contextual lookups require secondary lookups in the GSUB table for the actual substitution. makeotf creates a single secondary lookup for all three primary lookups in the pres feature. It sets the useMarkFilteringSet bit, and uses the mark filtering set of the first primary lookup, so it looks at above base marks. However, the glyph that actually needs replacing in the last primary lookup is a below-base mark. Since the lookup requests filtering for above base marks, the shaping engine should skip the below-base mark and not replace it.

In reality, the current versions of the major browsers (Safari 11.0, Firefox 56.0, Chrome 61.0.3163.100, Edge 40.15063.0.0) render text as intended for the font, apparently ignoring the mark filtering set of the secondary lookup. However, native apps on Android version 6.0 or older render text incorrectly, the glyph uni102F remains visible, which indicates that the mark filtering set is respected and the final substitution is not made.

The text used for testing is U+1004 U+103A U+1039 U+100F U+103C U+102F. The intended shaping behavior is:
– reordering: uni103C uni100F uni1004 uni103A uni1039 uni102F
– rphf feature: uni103C uni100F uni1004103A1039 uni102F
– pres feature, first lookup: uni103C.tt uni100F uni1004103A1039 uni102F
– pres feature, second lookup: uni103C102F.tt uni100F uni1004103A1039 uni102F
– pres feature, third lookup: uni103C102F.tt uni100F uni1004103A1039 __.invisible

It’s not clear why makeotf sets any lookup flag at all for the secondary lookup. By the time this lookup is reached, the primary lookup has decided that the substitution should be made, so no more filtering seems necessary or appropriate. But if for some reason the filtering is needed, then substitutions with different filtering settings must be kept in separate secondary lookups.

[AFDKO bug.zip](https://github.com/adobe-type-tools/afdko/files/1351159/AFDKO.bug.zip)

readroberts:
[...]

PR #538, opened 2018-08-09
[makeotfexe] generates incorrect secondary lookups with useMarkFilteringSet flag

readroberts:
[makeotfexe] generates incorrect secondary lookups with useMarkFilteringSet flag · Issue #196

Fix bug by adding check that lookup flag and mark filter sets match, before adding a new substitution rule to the current secondary lookup table.


c/makeotf/makeotf_lib/source/hotconv/GSUB.c
```diff
@@ -2229,7 +2229,7 @@ static Label addAnonRule(hotCtx g, GSUBCtx h, GNode *pMarked, unsigned nMarked,
 
         si = dnaINDEX(h->anonSubtable, i - 1);
         /* Don't need to match lkpFlag */
```
miguelsousa: Is this comment about not needing to match `lkpFlag` outdated now?

readroberts:
@miguelsousa @cjchapman Norbert's comment is misleading - his test case did have a ligature substitution, but that lookup did not have a problem. The bug showed up in the contextual single substitutions. In any case, the bug fix will apply to all the lookup types, so I think the current test is sufficient.

I do think the hotconv library should get bumped. However, I am not clear on how often we need to do this - for every change, or for every release, which may span several changes to makeotfexe/hotconv library. There is some overhead; as long as we not skip the name table Version string, all the ttx files in an expected_output directory need to be updated whenever either hotconv or makeotfexe versions are bumped.

The comment is indeed out of date- I will fix that immediately.

miguelsousa:
@readroberts I think it's good to bump the version every time that library is changed, even though we may not issue afdko releases after each change. I'm aware of the overhead this creates, and I've been meaning to add an option to the runner to address it. I'll take this opportunity to do it, and finish the PR.

PR #552, opened 2018-08-13
[stemhist] Integration tests & fixes

miguelsousa:


khaledhosny:
Note that it is still possible to build `autohintexe` from psautohint repo and use it as a drop-in replacement of the current one, without importing any of the changed Python code. At least in theory, I haven’t tried that in a while. Worth considering as a way to kill `autohint` earlier with little disruption.

khaledhosny:
I just tried that and it seems to work with the few fonts I threw on it.

miguelsousa:
@khaledhosny would you be opposed to including `autohintexe` in the `psautohint` wheel, at least for a period of time? Not having to build autohintexe from the afdko repository would allow us to nuke that old source code, and enable a progressive transition. Is this something that you can setup?

khaledhosny:
No objection, though I don’t know how to that (and the current build system I have for building `autohintexe` is not the most Windows friendly).

miguelsousa:
OK. For now I'm going to open an issue to track it.

miguelsousa:
done https://github.com/adobe-type-tools/psautohint/issues/79

PR #555, opened 2018-08-14
Add psautohint to the list of dependencies

miguelsousa:


.appveyor.yml
```diff
@@ -5,7 +5,6 @@ environment:
 
   matrix:
     - PYTHON_HOME: "C:\\Python27-x64"
-    - PYTHON_HOME: "C:\\Python27"
```
anthrotype: As far as i know, on Windows python 2.7 32-bit is much more widely used than the 64-bit version. the default download link on the python.org website is still the 32-bit one. Also, FontLab Studio 5 is a 32-bit application so it requires 32-bit python. I remember it was tricky to attempt installing both 32-bit and 64-bit pythons in the same machine as the installers would first uninstall the other for some reasons. Anyway I suggest you keep supporting the 32-bit version of 2.7 for Windows until you 2.7 EOL

miguelsousa: thanks for the feedback.
I had already changed it in 2a18fb852eaacac783d42e6ce691510732905cd8, because of a conflict between updating to VS2017 and pip installing psautohint.

Next step is enabling Python 3.6 on the CIs. All the tests already pass locally, and coverage has reached 50%, so it's about time to start making the move.

PR #568, opened 2018-08-17
cleanup spot

cjchapman:
* reformatted spot code using clang-format + hand-editing
* added cpplint of spot to CI


cjchapman:
The issues reported by Codacy are not new,  and are outside the scope of the level of cleanup I'm currently doing. My plan is to do another cleanup pass using `cppcheck` (which is what Codacy uses) sometime after I complete the current pass with `cpplint`.

PR #579, opened 2018-08-22
fix typedefs

cjchapman:
* modernized some typedefs to use `int32_t`, `int16_t`, etc.
* removed a lot of unused code from `supportpublictypes.h`


cjchapman:
@readroberts I'd appreciate it if you can confirm that `supportpublictypes.h` is OK, as the changes are extensive.

PR #593, opened 2018-08-28
Python 3 support

miguelsousa:
All is working except:
* ~~`makeinstancesufo_test.py` is failing on Py3 because of a mismatch in the sorting of anchors in **glif** files. There could be other hidden issues.~~
* we can't build a py36_amd64 wheel for Windows because:
  * `autohintexe` can only be built as 32bit
  * the Visual Studio files are missing x64 Release|Debug configurations


cjchapman:
The mismatch in the sorting of anchors in glif files is due to a bug in fontMath's `_anchorTree()` function which is using a `dict` when it should be using an `OrderedDict`. (Note that in Python 3.6 and later, `dict` acts like an `OrderedDict`.)

PR #596, opened 2018-09-05
Use autohintexe from psautohint when available

khaledhosny:


miguelsousa:
@khaledhosny thanks. I'm going to leave this unmerged until after a new stable afdko version (2.8.0) is released.

khaledhosny:
No idea why or where the CI is failing.

PR #606, opened 2018-09-11
Remove autohintexe

miguelsousa:
We currently get this tool via psautohint v1.8.1


Arnie97:
Hi, I got an error message after the 2.8.2 release:
> Please re-install the FDK. The path to the program 'autohintexe' is not in the environment variable PATH.

This seems to be a Windows-specific issue -- the same release works well on my Linux box. Neither platform has `autohintexe` in its `$PATH`.

miguelsousa:
That message seems to be coming from this line, https://github.com/adobe-type-tools/afdko/blob/develop/python/afdko/autohint.py#L556

The Windows CI is not failing, so I'm not sure how to reproduce your problem.
Did you have a previous AFDKO installed?

Arnie97:
Well, this was caused by spaces in the `psautohint.AUTOHINTEXE` path...

miguelsousa:
@Arnie97 that shouldn't happen. I'd like to fix this. Where exactly are the spaces causing the problem?

PR #630, opened 2018-09-28
[makeotfexe] Clarify path resolution of include statements

readroberts:
Fix issue that feature files referenced from a UFO font won't work in makeotfexe.

Note that for each include statement encountered, the current logic will try in order:
- resolve path relative to main feature file
- if main feature file is in a UFO font, resolve path relative to UFO font directory
- resolve path relative to including file.

This does mean that a font project will build even if it mixes all three models in different include files. The problem is that the is no way to know which model is intended until the first relative include path is encountered that will will work with only one of the models, and it even then it would be possible to draw the wrong conclusion. Ultimately, the most reliable approach would be to provide options so that a user could force one method or another. I think that scenarios where this will be a problem are sufficiently rare so as to not to be worth dealing with.

This issue touches on two problems. The features file spec says that feature include files are relative to the parent including files, but several implementations assume that relative paths are all relative to the main feature file. The second problem is that the UFO spec says that for UFO fonts, a relative include file path is always relative to the UFO font directory. In the discussion thread for this issue, it was agreed that for non-UFO fonts, the default is to resolve relative include paths relative to the main feature file, and relative to the UFO font directory for UFO fonts. Also, for legacy project support, if an include file cannot be found, makeotfexe should also look relative to including file.


miguelsousa:
@readroberts no update to the FEA spec?

PR #631, opened 2018-09-28
[makeotfexe] garbage in subtable overflow message 

readroberts:
The original report was because overflows reported for subtables had unreadable feature tags for two cases: the subtable belonged to a stand-alone named lookup, or to an anonymous lookup created by a chaining contextual rule. This is equally a problem for all the other error reports as well.

First, I standardized the *many* different overflow reports to a single function for each of the GPOS and GSUB tables.

The I standardized the feature name and lookup names to a single string, 'error_id_text', that is storied in the hotCtx structure, and used it in all the error reports, not just the overflow error reports. All error reports now identify the feature they come from. For both the overflow and other errors, some reports reported this, some didn't.

I then added logic so that this ID text uses the parent feature tag for anonymous lookups, and uses the lookup names for named lookups.

An issue is that some error and overflow reports are encountered only in the GPOSWrite and GSUBWrite phases, when only the subtable info is available, and the context where these are filled no longer exists. In order to allow reporting the feature ( and named lookup), I added a field to the standard Subtable structure to hold the 'error_id_text' string. This field is set when the GPOS and GSUB rules are being filled from the feature file, and is then available for use during the write phase.

Unfortunately, I could not test many of the coverage offsets reports. Because of how makeotfexe packs the subtables, it is not possible force a coverage table overflow without first encountering the check that one of the lookup subtables overflows the offset limit. I did not remove these checks, as they could be useful if we change how makeotfexe orders the subtables.

In addition to fixing the original problem, I fixed two other issues that surfaced when writing the tests:

1) The test files for the GPOS mark_to_base tests were failing under Linux because of an old makeotfexe bug in defining anchor lists. The logic defined a pointer to the previous anchor in an existing list of anchors, and then added a new anchor to the list. The list has a pre-allocated set of anchors; when a new anchor exceeds the length of this list, the entire list may be re-allocated and moved. If the list is moved, this invalidates in pointers into the prior location. Under Linux, this caused the pointer to the previous anchor to be bad, causing a crash.

2)  The logic for finding an included feature file failed if the last path separator, before the base file name, differed from the default for the OS. In the file test_makeotfexe.py, the paths to the feature file included a directory, with is perforce OS specific.

Fixed by changing the logic so that makeotfexe would work irrespective of the OS path separator used, or with mixed directory separator characters.


c/makeotf/makeotf_lib/source/hotconv/GPOS.c
```diff
@@ -770,16 +774,12 @@ void GPOSLookupEnd(hotCtx g, Tag feature) {
 
     switch (h->new.lkpType) {
         case GPOSSingle:
-            if (h->new.single.cnt == 0) {
-                hotMsg(g, hotFATAL, "Empty GPOS Single lookup");
-            }
+            /* Removed test that at least 1 GPOSSingle rule exists. Can't get here if that is the case */
```
cjchapman: I suggested changing comments like:
```/* Removed test that at least 1 GPOSSingle rule exists. Can't get here if that is the case */```
to something like:
```/* No need to test that at least 1 GPOSSingle rule exists. Can't get here if that is the case */```

cjchapman:
@readroberts I approved the changes that addressed the concerns I raised, but it looks like there are some build and merge issues that still need to be addressed.

PR #634, opened 2018-10-02
testing continuous integration

cjchapman:
this is just a test of the continuous integration systems, please do not merge


cjchapman:
@anthrotype This is the PR in which @miguelsousa and I are experimenting, as I mentioned on Gitter.

anthrotype:
try updating to the latest cibuildwheel from upstream (0.10.0). It worked for me on trufont/uharfbuzz repo

miguelsousa:
@anthrotype I don't see why that would make any difference, but I'll try. One noteworthy difference between afdko and uharfbuzz is that we build py27 wheels.

anthrotype:
why do you think it's related to wheel==0.32? If I look at the travis build logs, the wheel seems to be compiled successfully (or at least it doesn't error out), but then some assertions fail in the tests, as the generated output is different from the expected one..

anthrotype:
oh I see now:
https://travis-ci.org/adobe-type-tools/afdko/jobs/436340698#L315

delocate is failing with wheel==0.32.0

you need to pin wheel==0.31.1 in the `pyproject.toml`, under `[build-system] requires`. That should fix it. I did the same with uharfbuzz.

miguelsousa:
@anthrotype thanks, that's the bit I was missing.
@cjchapman closing this PR in favor of #637.

cjchapman:
@miguelsousa OK. I went ahead and deleting the test branch too.

Issue #402, opened 2018-05-27
[makeInstancesUFO] move to ufoProcessor?


LettError:
The designspace reader/writer in `mutatorMath.ufo` has been succeeded by [ufoProcessor](https://github.com/LettError/ufoProcessor) and [fontTools.designspaceLib](https://github.com/fonttools/fonttools/tree/master/Lib/fontTools/designspaceLib). 

* `designspaceLib` is the reader and writer
* `ufoProcessor` is a subclass that can generate UFOs with both mutatorMath and varlib.model if necessary. The varlib.model math has some advantages with non-aligned masters, but it can not extrapolate. 

The designspace format has been upped to `4` to reflect some changes in the rule element. I have no plans to update `mutatorMath.ufo`.

miguelsousa:
Thanks for the heads-up. I'll look into it.

miguelsousa:
I've published **ufoProcessor** to PyPI https://pypi.org/project/ufoProcessor/

LettError:
Thank you @miguelsousa!

PR #645, opened 2018-10-11
Update ufoLib to v3.0.0

miguelsousa:


anthrotype:
:crossed_fingers:

anthrotype:
ok that failure is a known one, but believe me it's a _feature_ ;)
https://travis-ci.org/adobe-type-tools/afdko/jobs/440211784#L2126

basically makeinstancesufo or mutatorMath (not sure) it attempting to write a UFO whose parent directory doesn't exist. Previously ufoLib would silently create all the directory tree (as in `mkdir -p`), now it fails with an error ("Cannot write to ...; the directory does not exist"), which is the right thing to do. 

See https://github.com/unified-font-object/ufoLib/issues/60

anthrotype:
this other failure is because checkoutlinesufo.py is modifying the `UFOWriter.layerContents` dictionary using a `bytes` string as the key (line 139, the global variable `PROCD_GLYPHS_LAYER_NAME`).
It should use unicode string instead, because the pyfilesystem2 API only accepts unicode strings for path names.

https://travis-ci.org/adobe-type-tools/afdko/jobs/440211784#L1518

miguelsousa:
I was going to submit a patch to **mutatorMath** for the directory tree creation problem, but I think a better move for **afdko** is to resume the work on #402.

anthrotype:
that sounds like a better plan. but ufoProcessor will have the same problem

PR #646, opened 2018-10-11
Fix AFDKO issue#635

nozoka:
This patch adds a new member in struct cstrCtx_ to hold warning messages per glyph only.


miguelsousa:
@nozoka I've folded your changes into PR #649 as I wanted to have a test case that captures the previous and the new version of the warning messages.

nozoka:
Thank you!

PR #654, opened 2018-10-22
[autohint] Enclose the AUTOHINTEXE path in quotes

Arnie97:


miguelsousa:
@Arnie97 thanks, the patch looks good. There has been a [GitHub incident](https://blog.github.com/2018-10-21-october21-incident-report/) that impacts this PR. We're waiting for the status to go green before merging.

python/afdko/autohint.py
```diff
@@ -1237,7 +1237,7 @@ def hintFile(options):
 		else:
 			if os.path.exists(tempBezNew):
 				os.remove(tempBezNew)
-			command = AUTOHINTEXE + " %s%s%s%s -s %s -f \"%s\" \"%s\"" % (verboseArg, suppressEditArg, supressHintSubArg, decimalArg, NEWBEZ_SUFFIX, tempFI, tempBez)
+			command = '"%s" %s%s%s%s -s %s -f "%s" "%s"' % (AUTOHINTEXE, verboseArg, suppressEditArg, supressHintSubArg, decimalArg, NEWBEZ_SUFFIX, tempFI, tempBez)
```
anthrotype: if you called the subprocess with a list of strings instead a single string, then subprocess would take care of the quoting for you

miguelsousa: `autohint.py` is on the chopping block

miguelsousa: and `fdkutils.runShellCmd` is used in many places; any changes to it need to be done carefully.

Issue #618, opened 2018-09-18
[tx] FontBBox values are not recalculated


miguelsousa:
- [x] `t1` format — PR #656
- [x] `cff` format — PR #659
- [x] `afm` format — PR #660

[Test font](https://github.com/adobe-type-tools/afdko/files/2394670/font.pfa.zip)
The original **FontBBox** values `{-151 10 531 820}` are incorrect.

Test commands:
`tx -t1 font.pfa font2.pfa`
`tx -t1 -g 0-1 font.pfa font3.pfa`

Expected FontBBox results:
* **font2.pfa**: `{25 -13 560 739}`
* **font3.pfa**: `{45 -13 560 669}`

miguelsousa:
I've created branch `fix-tx-bbox-bug618` with these tests.

PR #656, opened 2018-10-25
[tx] recalculate font bounding box when writing type 1 fonts

cjchapman:
part of fix for #618


cjchapman:
@readroberts When you review this, I'd especially like your confirmation that the calls from the callback methods in `t1write.c` to the metrics methods are in the correct places?  For example, in `glyphWidth`, I make the call to `h->glyph_metrics.cb.width` *after* a few `return` statements. Is that correct? Thanks!

readroberts:
I reviewed all the references to' h->glyph_metrics' in t1write.c, and this all looks good to me. In all the glyph callbacks, the early returns are for fatal errors, so it doesn't matter whether you call them early or late. If you want to be totally consistent, you could move the metrics call-back to the same relative location in each function, either at the top of the functions, or before the '/* choose format */' comment (glyphLine is currently different than glyphMove and glyphCurve), but this makes no difference in functionality.

cjchapman:
@readroberts Thanks for the feedback. We've got other branches waiting on this one, so I think we'll go ahead with it as is, and I'll plan on submitting another PR later to make the callbacks more consistent as you suggested.

PR #664, opened 2018-11-02
cleaned up xcode build warnings

cjchapman:
(mostly a port of fixes from tx.c)

### DIFF
--- a/c/mergefonts/source/mergeFonts.c
+++ b/c/mergefonts/source/mergeFonts.c
@@ -514,7 +514,7 @@ struct txCtx_ {
         Stream dbg;
         long flags; /* Library flags */
         int lenIV;
-        int fd;               /* -decid target fd */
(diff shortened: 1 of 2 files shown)


cjchapman:
I'd definitely like to consolidate the shared code, preferably sooner rather than later, but not as soon as this PR. Do we already have an issue about consolidating the code? Let me know if not, and I can create one. It's a bit more work than a simple clean up PR like this.

Also, I've got another set of changes to remove some more unused code that I'd like to commit after we merge this PR (removing some other files, with some implications for these files).

miguelsousa:
There's no open issue about code consolidation.

cjchapman:
There is now: https://github.com/adobe-type-tools/afdko/issues/665

cjchapman:
I had to redeliver the AppVeyor delivery just now to kick off its PR review.

Issue #675, opened 2018-11-07
Cannot install from source using Python3


frankrolf:
I solved my problem mentioned in #674 (by running the setup.py file using Python 2).
Not sure this is relevant, but Python 3 does not allow installing through setup.py:

```
Traceback (most recent call last):
  File "setup.py", line 274, in <module>
    main()
  File "setup.py", line 268, in main
    'install': InstallPlatlib,
  File "/usr/local/lib/python3.6/site-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/usr/local/lib/python3.6/site-packages/setuptools/command/install.py", line 67, in run
    self.do_egg_install()
  File "/usr/local/lib/python3.6/site-packages/setuptools/command/install.py", line 117, in do_egg_install
    cmd.run()
  File "/usr/local/lib/python3.6/site-packages/setuptools/command/easy_install.py", line 412, in run
    self.easy_install(spec, not self.no_deps)
  File "/usr/local/lib/python3.6/site-packages/setuptools/command/easy_install.py", line 654, in easy_install
    return self.install_item(None, spec, tmpdir, deps, True)
  File "/usr/local/lib/python3.6/site-packages/setuptools/command/easy_install.py", line 701, in install_item
    self.process_distribution(spec, dist, deps)
  File "/usr/local/lib/python3.6/site-packages/setuptools/command/easy_install.py", line 726, in process_distribution
    self.install_egg_scripts(dist)
  File "/usr/local/lib/python3.6/site-packages/setuptools/command/easy_install.py", line 600, in install_egg_scripts
    dist.get_metadata('scripts/' + script_name)
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1405, in get_metadata
    return value.decode('utf-8') if six.PY3 else value
[...]

PR #677, opened 2018-11-07
update instructions for installing from source

cjchapman:
resolves https://github.com/adobe-type-tools/afdko/issues/675
[no CI]

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,9 +125,10 @@ In order to build afdko from source get the files from the [afdko github
 repository](https://github.com/adobe-type-tools/afdko), cd to the
 top-level directory of the afdko, and use the `setup.py` script:
 
-    python setup.py install
+    python setup.py bdist_wheel
+    pip install dist/*.whl
 
-And to be able to run this install command, you must first have
+And to be able to run these commands, you must first have
 installed the development tools for your platform.
 
 On the Mac, install these with:


frankrolf:
I was about to ask “Are we really removing the simple `python setup.py install` – which works for Python 2”, but then I realized we probably don’t want to encourage Python 2 use anyway.

cjchapman:
The wheel-based instructions work for both Python 2 & 3.

The basic install instructions don't work because they're meant for Python-only installation, not the sort of hybrid C-binaries + Python installation that we're doing. At least, that's my understanding of the problem.

anthrotype:
the only thing really missing from this wheel-based setup is the ability to pip install in --editable mode, which is the same as doing `python setup.py develop`. But this only impacts the few core afdko developers that need to work on the sources as they modify them, and they can use alternative methods like exporting a PYTHONPATH, or using .pth file or similar hacks

anthrotype:
i'd suggest something like:
```
ensure the `wheel` module is installed (`pip install wheel`), then cd to the top-level directory of the afdko, and run:
    pip install .
```

anthrotype:
bonus point, with an additional note saying that `pip install -e .` doesn't work for the reasons explained above.

cjchapman:
@miguelsousa and @anthrotype I don't understand what you guys are talking about and I'm trying to focus on the refactoring of tx, rotatefont, & mergefonts right now, so I'd be grateful if one of you could update the readme with whatever wording you think is appropriate.

miguelsousa:
Will do.

PR #679, opened 2018-11-08
[tx] consolidate code from tx, rotatefont, & merge fonts into tx_shared library

### LINKED ISSUE #665: consolidate duplicate code from tx.c, rotateFont.c, and mergeFonts.c
Currently we have approximately 80% duplicate code between `tx.c`, `rotateFont.c`, and `mergeFonts.c` which is a maintenance issue as we're mainly trying to keep fixes in sync across the files. So, we need to consolidate as much of this code as we can into one place.

cjchapman:
resolves #665

Thanks to @readroberts for the help and moral support while I sorted out the various solution files, project files, etc.


readroberts:
@cjchapman Work to date looks good to me. I do have a question, though - how come you didn't also move all the mode help functions, and the doFile()/doMultiFileSet()/doSingleFileSet()/doAutoFileSet() functions into the  shared library? This all look like identical to me.

cjchapman:
@readroberts They are identical, but they use `#include`'d data that is not identical.

Issue #685, opened 2018-11-21
[tx] subroutinizer 64K limit


blueshade7:
Subroutinizing the attached CFF2 font generates an invalid charstring.
[uni7e7c.otf.zip](https://github.com/adobe-type-tools/afdko/files/2605297/uni7e7c.otf.zip)
```
% cp uni7e7c.otf uni7e7c_subr.otf
% tx -cff2 -no_opt +S uni7e7c.otf uni7e7c_subr.cff2
tx: --- uni7e7c.otf
tx: (cfr) Warning: CharString of GID 1 is 68301 bytes long. CharStrings longer than 65535 bytes might not be supported by some implementations.
tx: (cfw) unhinted <gid00001>
tx: (cfw) subr stack depth exceeded (reduced)
% sfntedit -a CFF2=uni7e7c_subr.cff2 uni7e7c_subr.otf
% tx -3 uni7e7c_subr.otf
tx: --- uni7e7c_subr.otf
```
***SNIP***
```
  774.11 -405.42 774.77 -405.41 774.77 -407.93 curve
  774.77 -407.94 774.35 -407.95 772.56 -407.96 curve
tx: (cfr) (t2c) invalid operator <gid00001>
tx: (cfr) charstring parse error
% 
```

blueshade7:
length > 64K appears to be overflowing unsigned short offsets in subroutinizer

PR #687, opened 2018-11-29
fix issue #685 [tx] subroutinizer 64K limit

blueshade7:
use "unsigned" in place of "unsigned short" as type of offsets and lengths
verified the fix using the attached CFF2 font uni7e7c.otf


blueshade7:
OK I re-worked it to use explicit sized types.

cjchapman:
@blueshade7 Thanks!

PR #692, opened 2018-12-12
Fix checkoutlinesufo hashes when writing to default layer

### LINKED ISSUE #349: [checkoutlinesufo] Hash is stale on first run when writing to default layer
Later in the workflow this bug will cause `makeotf` to fail building fonts, emitting messages like this:
`src glyph Aacute is newer than processed layer. programName: CheckOutlines`

Steps to reproduce the problem:
1. Start with a UFO without a **processedHashMap** file. The UFO must contain a glyph that has overlaps or components.
2. Run the command `checkoutlinesufo -e -w -q font.ufo`

**Result:** Glyphs that had overlaps and/or components will now be flat; file processedHashMap will be created but the hashes of those glyphs are stale, as they represent the glyphs' original state.

**Workaround:** Run the command again; this will update those glyphs' hash.

I'll add a test case for this once the `integration-tests` branch is merged.

miguelsousa:
Fixes #349


cjchapman:
Note to future GitHub historians: all the `force-pushed` transactions above were the result of me being really spaced out at the end of an exhausting day. Thanks to Miguel for his patience.

frankrolf:
May the Force be with you

PR #698, opened 2018-12-16
[makeotfexe] Correct "glyph not in font" error

### LINKED ISSUE #492: [makeotfexe] FEA token reported in syntax error is off-by-one
When reporting a FEA syntax error triggered by finding a glyph name that is included in a glyph class but not present in the font, `makeotfexe` references the next token instead of the name of the missing glyph.
* If the missing glyph **is the last in the class**, the syntax error will reference `]` (a bracket).
* If the missing glyph **is NOT the last in the class**, the syntax error will reference the name of the glyph that comes after the name of the missing glyph.

Test case:
```
makeotfexe -f t1pfa.pfa -ff features.fea
```

Result:
```
syntax error at "a" [features.fea 2]
```

Expected result:
```
syntax error at "glyph_not_found" [features.fea 2]
```
[bug492.zip](https://github.com/adobe-type-tools/afdko/files/2204416/bug492.zip)

miguelsousa:
Fixes #492


miguelsousa:
Thanks @cjchapman. I'll wait for @readroberts review before merging.

Issue #721, opened 2019-01-31
[tx] Fails with subroutinized CFF2 font


readroberts:
The function t2cstr.c::refill() has a test that returns error if when reading at the requested offset returns 0 bytes read because the offset is at the end of the file. 'refill()' is called at the start and end of each callsubr or callgsubr command, and to get get more bytes from the charstring. In CFF fonts, this cannot happen without it being an error, as when refill() is called, there is always at least an 'endchar' operator left.  In CFF2 fonts, this condition can happen normally. If the subr being called is the last item in the file, and the last item in this subr is another subr call, the current offset is set to the end of the file.

PR #722, opened 2019-01-31
Fix subr cff2 bug

readroberts:
Fixes AFDKO issue "[tx] Fails with subroutinized CFF2 font #721"


miguelsousa:
This PR is now green. The changes look fine but the test font doesn't need to be this big. I'll make a much smaller one and then we can merge.

PR #727, opened 2019-02-06
Fix wrong entries in Standard Apple Glyph Ordering

khaledhosny:
Three entries in that file are different from the table in [Apple spec](https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6post.html).

I have no idea what is the source of the difference (the file has them wrong since first commit), but it causes makeotf to fail compiling feature files to TTF fonts that reference glyphs using the Apple names.


khaledhosny:
BTW, how do I run the test suite locally? (I added a test, but I couldn’t figure a way to run the tests).

miguelsousa:
> how do I run the test suite locally?

- cd to afdko directory
- run `pip install -r requirements_dev.txt`
- run `pytest -v`

These steps assume that you have the afdko installed from source. If you don't, run `pip install .`

PR #732, opened 2019-02-07
Fix extension bug 321

readroberts:
[makeotfexe] Fix bug where useExtension is not applied to GSUB anonymous lookups.

Fixes issue 321, where useExtension is not applied to anonymous lookups from GSUB contextual substitutions

PR #734, opened 2019-02-07
[makeotfexe] Fix bug where useExtension is not applied to anonymous lookups.

### LINKED ISSUE #321: [makeotfexe] 'lookup useExtension' does not work right with contextual substitution lookups
When a  contextual lookup is made into an extension lookup, the actual contextual rules do go in extension subtables. However, the substitution rules that are applied by the contextual rules are accumulated in a separate lookup, whose subtables are NOT made into extension subtables. The problem code has been identified, but it will be a few days before the fix makes into a release.  This bug was found by Edgar Walthert, who is the first to push on this boundary.

readroberts:
Fix bug where useExtension is not applied to anonymous lookups produced by GSUB contextual lookups. Fixes #321.


miguelsousa:
Ugh! This PR has the same **Merge branch...** stuff as #732. I'll just fix it myself to not waste everyone's time any further.

PR #735, opened 2019-02-07
[makeotfexe] Fix offset overflow bug.

### LINKED ISSUE #731: [makeotfexe] Fails to detect offset overflow error
The offset from a lookup record to a subtable is measured from start of the LookupList to the start of the subtable. However, when filling the subtables, you can't yet know how large the LookupList is. When makeotfexe calculates the offset to a lookup subtable, it actually calculates it only from the end of the lookup list, and the checks for an offset overflow are applied to this value. The final value written to the font, however, has the length of the lookup list added: thus an offset which is just a bit smaller the 64K can get pushed over the edge. These are most likely to happen in the aalt feature, as these lookups are almost always first in the LookupList,  and (for arcane reasons) the subtables are sorted to the end of the subtable list, so the offsets from the aalt lookup tables to their subtables is the greatest for any of the subtables. You can reduce this effect by using useExtension with the aalt feature.  This bug is also more likely to happen when  many contextual rules are used, as makeotf makes a separate subtable for each rule, which makes the lookup table which contains the rules much larger, hence the size of the LookUpList is greater. Thanks to Paul de Laan and Edgar Walthert for finding this bug.

readroberts:
Fixes #731

Check for offset overflow when final subtable offset value is set by adding the the size of the LookupList by to the current subtable offset.


miguelsousa:
This is a redo of #733

Issue #702, opened 2018-12-18
[makeotf] documentation contains no information about glyphs.com.adobe.type.processedglyphs layer


frankrolf:
I know about the `glyphs.com.adobe.type.processedglyphs` layer, which is created by `checkoutlinesUFO` and `(ps)autohint`, and which `makeotf` prefers when building a font. However, there is no mention of this layer in the documentation at `makeotf -h`.

Additionally, the tabbing in the documentation is a little hard to read (not sure that’s easy to control):
![screen shot 2018-12-18 at 16 38 17](https://user-images.githubusercontent.com/2049645/50164582-64979c00-02e3-11e9-8c81-9f0496e50425.png)

miguelsousa:
@frankrolf please submit a fix

frankrolf:
Will do!

miguelsousa:
Fixed via #738

PR #738, opened 2019-02-10
[makeotf] Update and re-format documentation

frankrolf:
The only major addition is the acknowledgement of the processed glyphs layer, as pointed out in #702.
I fixed some typos and aligned the grammar in a few cases. Finally, I made the tabbing/whitespace more consistent and readable.

Minor remark:
I see an issue with keeping the file `docs/MakeOTFUserGuide.pdf` up to date – I think we could consider either building this file automatically, or think of replacing it with a sphinx-style documentation.


miguelsousa:
I'd be more inclined to convert **docs/MakeOTFUserGuide.pdf** to markdown, to keep things simple in terms of necessary setup.

Issue #594, opened 2018-08-31
[spot] Replace hacky temp file solution


miguelsousa:
https://github.com/adobe-type-tools/afdko/blob/develop/c/spot/source/GPOS.c#L4667

PR #739, opened 2019-02-10
[makeotf] Use FontTools to copy font tables

khaledhosny:
The `sfntedit` tool creates temporary files with fixed names in the current working directory, which breaks running multiple `makeotf` instances in parallel (e.g. from a makefile). Using FontTools fixes this limitation.


miguelsousa:
`spot` is another tool with a less-than-ideal temp file solution #594

PR #749, opened 2019-03-01
[makeotfexe] Fixed detection of offset overflow to a FeatureParam, issue 746

readroberts:
The main issue is that the variable holding the offset was unsigned short, and needed to be a unsigned long. Check all other offsets checked this way to make sure than this didn't happen elsewhere.

Also fixed a bug where makeotf doesn't notice a non-zero return status from makeotfexe.

I did not yet check in a test case: I have one, derived from the fonts supplied by Paul De Laan, but first want to make sure I have their permission to use the features files. I can build a test font de novo, but it would take several hours.


BoldMonday:
FWIW – I sent files by email with permission to use them for testing.

miguelsousa:
Thanks for the files Paul.

miguelsousa:
While working on putting test cases in place, it became clear that the changes to makeotf and fdkutils have no effect. I'm therefore going to exclude them.

PR #754, opened 2019-03-05
Simplify setup.py a bit

khaledhosny:
get_executable_dir() is only used in one place to decide which build script to use.


cjchapman:
@khaledhosny Let me also add my thanks! 😄

PR #760, opened 2019-03-05
[makeotf] features file may have a suffix

frankrolf:
Minor (undocumented) feature of `makeotf`: the default `features` file may have a suffix (which helps with detecting syntax coloring).


frankrolf:
That’s done. Still some CI seems to show up above.

frankrolf:
Thank you!

Issue #160, opened 2016-12-21
[sfntedit] Ignore missing tables and continue


twardoch:
@readroberts Currently, when I use `sfntedit` to extract or delete a series of tables, and specify some tables that don't exist in the font (e.g. I say `sfntedit -d GPOS,GSUB font.ttf` and the font does not have `GSUB` but only `GPOS`), `sfntedit` issues an error `sfntedit [FATAL]: table missing (GSUB)` and does not perform the operation at all. It would be useful to have an option `-i` (ignore errors) that would ignore these situations and still perform the operations on the other tables which do exist. I don't know enough C to patch `sfntedit` myself but I think it should be a relatively minor but useful thing.

PR #769, opened 2019-03-13
[sfntedit] Fix sfntedit bug 160

readroberts:
Made changes to continue if a table is missing (a very small change), and to not change the output file if no changes are made.

Added test case to assert that a warning is shown for the attempt to delete or extract a missing table, but that the tool does continue and operates on  another table.


readroberts:
@msousa @cjchapman  @frankrolf I added a new option partly because that was part of the request in issue #160, and no other comment was made, and partly because it seems like a good idea to me. When I use sfntedit, I want it to work entirely, or not all and to exit with an error. In my work, by far the most common case of a missing table is when I mis-type a tag, and I want a clear indication that I did something wrong. I see fixing this issue as a concession to a particular user's workflow for a rare case, in the spirit of allowing developers to work as they want. Do you indeed think  that a skipping missing tables without an error should be the default behavior?

miguelsousa:
Yes, no option. Instead of an error make it a warning.

readroberts:
Will do. Removing the option is an easy change. Can you confirm that in the case fo a missing table, sfntedit should with exit with no error code? I think this is appropriate for just a warning.

miguelsousa:
Yes, exit with `0` and issue a warning.

readroberts:
Note that the changed logic was force-pushed, so you will need to update your copy of this branch.

c/sfntedit/source/main.c
```diff
@@ -976,6 +985,9 @@ static void sfntCopy(void) {
         totalsum += checksum;
     }
 
+        if (!changed)
```
cjchapman: please fix indentation here

cjchapman: (otherwise the changes look OK to me)

readroberts:
Fixed white-space issue.

PR #777, opened 2019-03-21
Convert OpenType Feature File specification to Markdown

miguelsousa:
Continuation of #776


miguelsousa:
@frankrolf I'm done with my changes and also implemented [Read's suggestions](https://github.com/adobe-type-tools/afdko/pull/776#issuecomment-475329287).

Unless you have further tweaks, I think the only remaining item is the TOC.

frankrolf:
@miguelsousa Thank you for this. I think changing the TOC to match the system proposed in the document structure might not be possible – I tried it earlier.
As of this moment, the TOC is just a system of nested, numbered lists like this:

1. item one
2. item two
    1. item two_one
    2. item two_two
        1. item two_two_one
        2. item two_two_two
        2. item two_two_three
    3. item two_three
3. item three

The GitHub markdown parser decides to put lowercase Roman numerals first, then alphabetic indexing – opposite to the way the spec works at the moment. The source code for the TOC is purely numerical:

```
1. item one
2. item two
    1. item two_one
    2. item two_two
        1. item two_two_one
        2. item two_two_two
        2. item two_two_three
    3. item two_three
3. item three
```

I tried to match that code to what’s set up in the document itself, but it breaks the whole system.
I am glad to update the numbering structure within the document to match whatever GitHub is proposing, but I think it may be different in another markdown interpretation.

miguelsousa:
Thanks, I'll look into it. I'd like to try some stuff.

PR #792, opened 2019-05-01
[buildcff2vf] Major rewrite for sparse font support

readroberts:
- support sparse source fonts
- subset full masters to source source fonts from subset definition file
- check and fix flat curve and closepath outline incompatibilities
- use argparse for arguments.


python/afdko/buildcff2vf.py
```diff
@@ -2,996 +2,498 @@
...
+    for li, line in enumerate(text_lines):
+        idx = line.find('#')
+        if idx >= 0:
+            line = line[:idx]
+        line = line.strip()
+        if not line:
+            continue
+        if line[0] == "(":
+            cur_key_list = []
+            location_list = eval(line)
```
cjchapman: Regarding the Codacy issue on this part of the code, if `line` is something like this:
```
line = "(42)"
```
then I suggest replacing the use of `eval()` with something like this:
```
location_list = int(line.strip("()"))
```

python/afdko/buildcff2vf.py
```diff
@@ -2,996 +2,498 @@
...
+        font_list = [src.font for src in designspace.sources]
+        default_font = designspace.sources[ds_data.base_idx].font
+        vf = deepcopy(default_font)
+        # We copy vf from default_font, because we use VF to hold
+        # merged arguments from each source font charstring - this alters
+        # the font, which we don't want to do to the default font.
+        do_compatibility(vf, font_list)
+
+    logger.progress("Building VF font...")
+    varFont, varModel, masterPaths = varLib.build(designspace, otfFinder)
```
cjchapman: Per Codacy, `masterPaths` is unused, so you could replace this line with something like:
```
 varFont, varModel, _ = varLib.build(designspace, otfFinder)
```
Assuming that the third value returned by `varLib.build()` is actually used elsewhere, otherwise just remove that from what's returned.

readroberts:
I will be pushing incremental updates to this branch for the next while - don't bother reviewing until I notify you that I am done.

python/afdko/buildcff2vf.py
```diff
@@ -1,998 +1,486 @@
...
+            if prev_moveto_coords2 != prv_coords[-2:]:
+                return False
+
+            # Add the closing line coords for this region
+            # so self._commands, then increment self.pt_index
+            # so that the current region op will get merged
+            # with the next default font moveto.
+            if cmd[0] == 'rrcurveto':
+                new_coords = self.make_curve_coords(prev_moveto_coords, False)
+            cmd[1].append(new_coords)
```
cjchapman: This line should be indented to the same level as the previous line, as `new_coords` is only assigned within the `if` block.

python/afdko/buildcff2vf.py
```diff
@@ -1,998 +1,486 @@
...
+
+    if not options.verbose:
+        level = PROGRESS_LEVEL
+        logging.basicConfig(level=level, format="%(message)s")
     else:
-        print(__usage__)
-        return
+        if options.verbose:
+            level = logging.INFO
+        logging.basicConfig(level=level)
```
cjchapman: The `level` might not be set as of this line or the next. Perhaps set `level` to some default before line 430 (i.e. the start of the `if...else...` block, or else sort this out via indentation.

python/afdko/buildcff2vf.py
readroberts: Right.

python/afdko/buildcff2vf.py
readroberts: Must have left this behind when editing logic. The issue is actually that the test 'if options.verbose' is redundant, as it alway true here.

readroberts:
@miguelsousa @cjchapman Requested changes made - please check. 

On another issue, the builds are currently failing because a) the build scripts are installing fontTools 3.41.2, and b) in version 3.41.2,  fontTools changed to keep glyph names in the 'post' table by default. This baffles me  - the requirements.txt file for the update-buildcff2vf branch require fontTools 3.41.0. How come the requirements file for the latest node on 'develop' is being used?

Of course, what is needed is to change buildcff2vf to work with 3.41.2. This requires some work - changing buildcff2vf to suppress glyph names rather than add them is trivial, but I think the fontTools has a bug in adding glyph names when the sources are CID fonts. Am looking into this.

python/afdko/buildcff2vf.py
cjchapman: OK, I see that this is fixed now, so resolving this conversation.

python/afdko/buildcff2vf.py
cjchapman: OK, I see that this is also fixed now, so resolving this conversation too.

anthrotype:
Oops, i forgot about CID... 😬
@readroberts can you please send a PR to fix that? Thank you

readroberts:
@anthrotype @msousa @cjchapman Turns out there is no problem in fontTools with CID fonts - I was just suffering from versionitis - I haven't fully grasped when I need to re-install fontTools to have the latest sources take effect,  when I have in installed in editable developer mode ( pip install -e .[...]).
Updated buildcff2vf to suppress postscript names, rather than add them, when appropriate.

PR #808, opened 2019-06-10
[makeinstancesufo] Preserve existing features.fea file

miguelsousa:
Every time UFO instances are regenerated all their files are first deleted. This is fine for most of the UFOs' contents but in certain situations‡ it's desired to have masters' `features.fea` files that are different from the instances' ones.
With this change, the contents of any existing `features.fea` file in the instances are preserved, as long as `<features copy="1"/>` is NOT used in any of the sources/masters definitions in the designspace file.

‡ when the UFO masters are used as source files for generating UFO instances as well as for generating variable fonts (e.g. [Source Sans](https://github.com/adobe-fonts/source-sans-pro/tree/master) and [Source Serif](https://github.com/adobe-fonts/source-serif-pro/tree/master))


frankrolf:
Confirmed working on my end. Thank you!

PR #815, opened 2019-06-19
Fix mark to base bug

readroberts:


miguelsousa:
@readroberts can the tests be for `makeotfexe` rather than `makeotf`? Tests against the former run faster as there's much less overhead.

miguelsousa:
Also, should `HOT_VERSION` get bumped?

readroberts:
@miguelsousa Agreed on both comments. Moved test to makeotfexe_test.py, bumped HOT_VERSION. Force-pushed clean branch.

PR #830, opened 2019-07-09
remove cube code

cjchapman:
* removed support for the obsolete cube operators
* made the three `txops.h` files more consistent


readroberts:
The only thing we actually use the library version numbers for is to be able to quickly compare many program and library version numbers between a afdko user and the latest versions. However, this is useful.

cjchapman:
OK. I've got it on my to-do list to update the relevant version numbers before the next release.

PR #835, opened 2019-07-11
[tx] fix SetNumMasters call to var_getIVSRegionIndices

cjchapman:
SetNumMasters was incorrectly passing the number of region indices to var_getIVSRegionIndices for the regionListCount. With this change it now passes the total region count for regionListCount.

Example symptoms of the bug:
```
tx: (cfr) inconsistent region indices detected in item variation store subtable 1
tx: (cfr) memory error
```


cjchapman:
@blueshade7 Yes, thanks for the reminder. I've amended the commit with a test case.

PR #837, opened 2019-07-11
[tests] remove xfail_py27_mac decorator

josh-hadley:
removed decorator definition and 1 instance


miguelsousa:
@josh-hadley can't `xfail_py3_win` be removed as well? The Windows build is 64bit.

josh-hadley:
I'm thinking let's defer that until after the C tools/Unicode paths stuff is better sorted (the tests using `xfail_py3_win` also seem to be path-related).

miguelsousa:
OK

PR #838, opened 2019-07-11
[makeotf] use absolute paths for temp files

### LINKED ISSUE #828: [makeotf] macOS 10.15 issue
I am currently poking around with the macOS 10.15 beta – in it, it no longer seems possible to run `makeotf` in its current configuration:

```
➜  Desktop makeotf -f test.ufo
tx: --- /Users/fg/Desktop/test.ufo
tx: (ufr) Warning: No version specified in source UFO font.
makeotf [Warning] Could not find default features file. Font will be built without any layout features.
makeotfexe [FATAL] file error <No such file or directory> [../../../var/folders/yl/cz8lx5qn6ps_4pw00l8gtm9c0000gp/T/tmpm9kopaya]
makeotf [Error] Failed to build output font file 'test-test.otf'.
```

however, this works:
```
➜  Desktop sudo makeotf -f test.ufo
Password:
tx: --- /Users/fg/Desktop/test.ufo
tx: (ufr) Warning: No version specified in source UFO font.
makeotf [Warning] Could not find default features file. Font will be built without any layout features.
makeotfexe [NOTE] Adding glyph .notdef, SID 0.
Built development mode font 'test-test.otf'.
```

I don’t know what mechanism expects a file in the var folder, I also don’t know which steps need to be taken to make `makeotf` behave as it used to, but I assume it’s good to know about this change.

cjchapman:
In macOS 10.15 it is not possible to reach temporary files in the `/var` folder using relative paths, hence this change in `makeotf.py` to using absolute paths for temporary files. See #828 for more details.

fixes #828

### DIFF
--- a/python/afdko/makeotf.py
+++ b/python/afdko/makeotf.py
@@ -2317,10 +2317,12 @@ def makeRelativePath(curDir, targetPath):
         return targetPath
 
 
-def makeRelativePaths(makeOTFParams):
+def adjustPaths(makeOTFParams):
     """
     Change file paths to be relative to fontDir,
     if possible, else to absolute paths.
+    (With the exception of fmndbFilePath and tempFontPath which are
+    always absolute paths.)
     """
     inputFilePath = getattr(makeOTFParams, kFileOptPrefix + kInputFont)
     fontDir = os.path.dirname(os.path.abspath(inputFilePath))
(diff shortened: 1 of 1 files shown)


cjchapman:
@miguelsousa I would be really surprised if `tempfile.mkstemp()` returned a path with a symlink that required `os.path.realpath()` to resolve.  For that matter, I only used `os.path.abspath()` to be very explicit that those are absolute paths, unlike the other ones which are relative. In my original [experiment](https://github.com/adobe-type-tools/afdko/issues/828#issuecomment-510293558) I did not use (and did not need) `os.path.abspath()`.

PR #849, opened 2019-07-16
[t2cstr] move _t2cCtx off the stack (and init with zeros)

cjchapman:
The structure `_t2cCtx` is about 600k in size and thus should be in the heap, not on the stack, hence this change. (Also added a `memset` to explicitly initialize it full of zeros.)


cjchapman:
I think Codacy is confused by the `DURING_EX` ... `HANDLER` ... `END_HANDLER` macros in this block:
```
    DURING_EX(h->err.env)

    retVal = t2Decode(h, offset);

    HANDLER
    retVal = Exception.Code;
    END_HANDLER
```

readroberts:
@cjchapman Can you point me to where the malloc'd _t2cCtx gets freed?

cjchapman:
@readroberts I can, now that I've amended the commit. :wink: Thanks for catching this.
(I also removed a couple of apparently superfluous lines.)

cjchapman:
I'm guessing it's the  `END_HANDLER` macro that's confusing Codacy into thinking that `free()` is a variable. 🤦‍♂

Issue #847, opened 2019-07-16
[makeotf] UnicodeDecodeError: 'ascii' codec can't decode byte


miguelsousa:
**makeotf** is unable to print a non-ASCII message coming from **makeotfexe**.

```
makeotfexe -f font.pfa -ff feat.fea 
makeotfexe [WARNING] unhinted <a>
makeotfexe [WARNING] glyph 'negative' has a negative advance width. Setting width to 0.
makeotfexe [WARNING] FontMenuNameDB file was not specified or not found. [SourceSansPro-Regular]
makeotfexe [FATAL] <SourceSansPro-Regular> invalid token (text was "?") [feat.fea 7]
```

```
makeotf -f font.pfa -ff feat.fea 
makeotf [Warning] Could not find FontMenuNameDB file. Font will be built with menu names derived from PostScript name.
makeotfexe [WARNING] glyph 'negative' has a negative advance width. Setting width to 0.
makeotfexe [WARNING] FontMenuNameDB file was not specified or not found. [SourceSansPro-Regular]
Traceback (most recent call last):
  File "makeotf", line 11, in <module>
    load_entry_point('afdko', 'console_scripts', 'makeotf')()
  File "makeotf.py", line 2720, in main
    runMakeOTF(makeOTFParams)
  File "makeotf.py", line 2596, in runMakeOTF
    fdkutils.runShellCmdLogging(commandString)
  File "fdkutils.py", line 122, in runShellCmdLogging
    print(tostr(output))
  File "fontTools/misc/py23.py", line 131, in tounicode
    return s.decode(encoding, errors)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xce in position 68: ordinal not in range(128)
```

[test_files.zip](https://github.com/adobe-type-tools/afdko/files/3395673/test_files.zip)

josh-hadley:
A data point: with current AFDKO (WIP, removing use of `py23` and others), I get:
```bash
makeotf -f font.pfa -ff feat.fea
makeotf [Warning] Could not find FontMenuNameDB file. Font will be built with menu names derived from PostScript name.
b"makeotfexe [WARNING] glyph 'negative' has a negative advance width. Setting width to 0."
b'makeotfexe [WARNING] FontMenuNameDB file was not specified or not found. [SourceSansPro-Regular]'
b'makeotfexe [FATAL] <SourceSansPro-Regular> invalid token (text was "\xce") [feat.fea 7]'
[...]

PR #850, opened 2019-07-16
[python] drop use of fontTools.misc.py23

josh-hadley:
- numerous modules
- also simplified `fdkutils.py` routines that do shell output (see [#847](https://github.com/adobe-type-tools/afdko/issues/847))

### DIFF
--- a/python/afdko/autohint.py
+++ b/python/afdko/autohint.py
@@ -441,7 +441,6 @@ import os
 import re
 import time
 import tempfile
-from fontTools.misc.py23 import open, tounicode
 from fontTools.ttLib import TTFont, getTableModule
 import plistlib
 import warnings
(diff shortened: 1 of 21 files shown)


python/afdko/autohint.py
```diff
@@ -1303,11 +1305,15 @@ def hintFile(options):
 					logMsg("No new hints. All glyphs were already hinted.")
 			else:
 				logMsg("No glyphs were hinted.")
+
 	if options.usePlistFile and (anyGlyphChanged or pListChanged):
 		#  save font plist file.
-		fontPlist.write(fontPlistFilePath)
+		with open(fontPlistFilePath, 'wb') as f:
+			plistlib.writePlist(fontPlist, f)
```
miguelsousa: use `dump()` instead?

josh-hadley: Eh. If we were planning to keep `autohint.py` current I guess that would be the right thing to do (actually, the right thing would be to use `fontTools.misc.plistlib`)

miguelsousa: only suggested it because you used `plistlib.load()` upstream

josh-hadley: Indeed. I'll change it to `dump()` in next round of changes.

PR #858, opened 2019-07-18
[spot] fix memory leaks and uninitialized memory

cjchapman:
Static analysis driven code cleanup:
* fixed some memory leaks
* fixed some uninitialized memory
* fixed some compiler warnings
* removed some dead and superfluous code

### DIFF
--- a/c/spot/source/GSUB.c
+++ b/c/spot/source/GSUB.c
@@ -2456,7 +2456,6 @@ static void decompileChainContext3(ChainContextSubstFormat3 *fmt) {
     unsigned short numSubRecs = fmt->SubstCount;
     SubstLookupRecord *slr = NULL;
     IntX needLookupRefs = (numSubRecs > 1);
-    IntX haveAnyInLineRepl = 0; /* have any sub rules that will print the replacement string, and hence new-line */
     IntX decompileable = 0;
     Lookup *lookup;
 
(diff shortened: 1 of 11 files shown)


readroberts:
Everything looks good, except for removing the "sfntTTCList()" function. I see this is never called, but it should be. The help documentation lists an option "-tttcf" that used to call it - I would suggest restoring the call. I don't see anything in the commits back to 2014 which suggest that this was removed deliberately.  

Miguel @miguelsousa  and @kenlunde , do you  remember anything about this?

kenlunde:
Nothing springs to mind.

readroberts:
If we restore the function, the help message should be updated: 'https://docs.microsoft.com/en-us/typography/opentype/spec/otff#collections" notes that a "TrueType Collection Font" should now be called an "OpenType Collection Font"

miguelsousa:
I don't recall ever using that option.

cjchapman:
@readroberts (et al.) I think that functionality is now handled by [dirDump()](https://github.com/adobe-type-tools/afdko/blob/7206cc35fbf0dbc27729e054c728a8d3320fb31c/c/spot/source/sfnt.c#L468-L503), e.g.:

```bash
$ spot -tttcf RgIt.ttc 
### [ttcf] (00000000)
TTCTag        =ttcf (74746366)
Version       =1.0 (00010000)
DirectoryCount=2
--- TableDirectory[index]=offset
[0]=00000014 [1]=00000100 
$ 
```

cjchapman:
Here's a hexdump of the TTC header from that font:
```bash
cchapman-macOS:input cchapman$ hexdump -C -n 32 RgIt.ttc 
00000000  74 74 63 66 00 01 00 00  00 00 00 02 00 00 00 14  |ttcf............|
00000010  00 00 01 00 4f 54 54 4f  00 0e 00 80 00 03 00 60  |....OTTO.......`|
```

cjchapman:
@readroberts Please see my comments above and my latest change and re-review, thanks.

PR #865, opened 2019-07-24
[svread] check size of token when doing compare

cjchapman:


cjchapman:
Yes, prior to this change if you do something like line 888:
```
        } else if (tokenEqualStr(tk, "font-family=")) {
```
and `tk.length` is, for example, `1`, the code will treat `tk.val="f"` as equal to `str="font-family="`.

cjchapman:
So, just to be clear, I made that change because, based on how `tokenEqualStr()` (the only caller of `tkncmp()`) is used, we definitely do not want a match if the token is shorter than the string we are comparing it against.

cjchapman:
I apologize for not explaining this in the commit message and/or the PR description.

blueshade7:
I noticed `uforead.c` has a function under the same name `tkncmp` with this definition:
```
static int tkncmp(token* tk, char* str) {
    size_t len = strlen(str);
    int retVal = 1;
    if ((tk != NULL) && (len == tk->length))
        retVal = strncmp(tk->val, str, tk->length);
    return retVal;
}
```
I believe it achieves the same goal so best to be consistent. Seems to me name `tkncmp` a bit misleading but may not be worth fixing for now. BTW, `uforead.c` has `tokenEqualStrN` for matching a prefix as a variant of `tokenEqualStr`. Another poorly named/designed function, take a look if curious.

cjchapman:
I have amended the commit accordingly. Thanks.

PR #877, opened 2019-08-01
[makeotf] prevent bash and python code execution

### LINKED ISSUE #780: [makeotf] bash code in style name is executed
This is another obscure one:
There was an instance where a student back-ticked their style name: 
```
Exercise3-`Semibold`
```

As a result, I noticed that `makeotf` seems to execute code in the style name while trying to compile the font:
```
makeotf [Note] Using features file at './features.fea'.
/bin/sh: Semibold: command not found
makeotfexe [WARNING] not in FontMenuNameDB [Exercise3-`Semibold`]
makeotfexe [FATAL] <Exercise3-`Semibold`> I can't find a Family name for this font !
makeotf [Error] Failed to build output font file 'Exercise3-`Semibold`.otf'.
```

I was curious what would happen if I set the style name of an UFO to a real bash command, and indeed it works:
```
		<key>postscriptFontName</key>
		<string>Exercise3-`ls`</string>

```
```
makeotf [Note] Using features file at './features.fea'.
makeotfexe [WARNING] not in FontMenuNameDB [Exercise3-`ls`]
makeotfexe [FATAL] file error <File name too long> [Exercise3-Exercise3-.otf
Exercise3-Exercise3-features.fea
font.ufo
fontinfo
kern.fea
mark.fea
markclasses.fea
mkmk.fea.otf
features.fea
font.ufo
fontinfo
kern.fea
mark.fea
markclasses.fea
mkmk.fea.otf
Exercise3-features.fea
makeotf [Error] Failed to build output font file 'Exercise3-`ls`.otf'.
```

I did not figure out how to include spaces, otherwise I’d have tried 
```
Exercise3-`rm -r font.ufo`
```

I am not sure how much of a real-world issue this is, but it seems a bit odd it’s possible for commands being executed this way.

josh-hadley:
- removed use of `subprocess` with `shell=True` which allows shell code execution (closes #780 )
- removed/replace use of Python `eval()` and `exec()` which allow Python code execution


python/afdko/makeotf.py
```diff
@@ -1223,17 +1234,13 @@ def getOptions(makeOTFParams, args):
...
-                    formatOK = 1
-                elif isinstance(fontRevision, float):
-                    _, minor = fontRevisionString.split(".")
-                    if len(minor) == 3:
-                        formatOK = 1
-            except ValueError:
-                pass
+            formatOK = False
+            frs_parts = fontRevisionString.split(".")
+            formatOK = frs_parts[0].isnumeric()
```
miguelsousa: What happens if someone uses `+1.000` or `-1.000` for revision number?

josh-hadley: a '+' or "-" before the numbers won't match either of the RegExes (for getting existing value from .fea file or user input with `-rev` CLI option), so `makeotf` won't modify the .fea file. I don't know what `makeotfexe` would do with a value like that if it's in a .fea file, that should probably be tested.

josh-hadley: More on this: `makeotf -rev -1.000` doesn't even get to that part of the code (the CLI arg parser doesn't recognize '-1.000' as an option). `makeotf -rev +1.000` trips the `if not formatOK` test. So I don't think this is a concern for `makeotf`. Again, probably should check what `makeotfexe` does when a .fea file contains values like that, but that's a different issue/PR, I think.

python/afdko/makeotf.py
```diff
@@ -2583,41 +2628,29 @@ def runMakeOTF(makeOTFParams):
...
     if makeOTFParams.verbose:
         print("makeotf [Note] Running %s with commands:" %
               os.path.basename('makeotfexe'))
-        print("   cd \"%s\"" % fontDir)
-        print("   %s" % commandString)
+        print(f'   cd "{fontDir}"')
+        print(f"   {params}")
 
-    fdkutils.runShellCmdLogging(commandString)
+    success = fdkutils.run_shell_command(params, suppress_output=True)
```
miguelsousa: On a closer look, the output cannot be suppressed because it suppresses all the errors and warnings issued by **makeotfexe**. Run this for example:

    makeotf -f makeotf_data/input/cidfont.ps

This is why I wanted to run a few commands first; the test suite doesn't tell the full story.

PR #882, opened 2019-08-05
[makeotfexe] port tx subroutinizer to makeotfexe

### LINKED ISSUE #331: [makeotfexe] Replace the subroutinizer with the one from tx

cjchapman:
resolves #331

### DIFF
--- a/c/makeotf/makeotf_lib/source/typecomp/subr.c
+++ b/c/makeotf/makeotf_lib/source/typecomp/subr.c
@@ -6,13 +6,15 @@
 /* Temporary debug control */
 #define DB_FONT 0   /* Font input charstrings */
 #define DB_INFS 0   /* Inferior subr selection */
-#define DB_OVLPS 0  /* Overlap handling */
+#define DB_ASSOC 0  /* Associate with subrs */
 #define DB_RELNS 0  /* Subr relationship building */
 #define DB_SELECT 0 /* Subr selection */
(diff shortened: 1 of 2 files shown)


c/makeotf/makeotf_lib/source/typecomp/subr.c
```diff
@@ -37,11 +39,17 @@
 
    The most challenging part of the process of subroutinization is finding and
    counting the repeated charstrings. This is achieved by first building a suffix
-   DAWG using the concatenation of all the charstrings from all the fonts as
-   input. Subseqeuntly the completed suffix DAWG is traversed in order to count
+   CDAWG using the concatenation of all the charstrings from all the fonts as
+   input. Subseqeuntly the completed suffix CDAWG is traversed in order to count
```
miguelsousa: Subseqeuntly -> Subsequently

cjchapman: I fixed that. I'm ignoring the Codacy warnings for now, as I want to keep this code reasonably close to the `tx` code. If I fix either, I want to fix both. (And yes, ultimately, I'd like to just have one shared library for this, but at least this is a step in that direction.)

cjchapman:
@blueshade7 That's correct. We use `makeotf` only for CFF. We use `buildcff2vf` for CFF2.

PR #898, opened 2019-08-12
[checkoutlinesufo] Fix support for non-UFO font formats

miguelsousa:


python/afdko/checkoutlinesufo.py
```diff
@@ -264,10 +242,17 @@ def _split_lines(self, arg, width):
         return [item for sublist in arg_rows for item in sublist] + ['']
 
 
+def _validate_path(path_str):
+    valid_path = os.path.abspath(os.path.realpath(path_str))
+    if not os.path.exists(valid_path):
+        raise argparse.ArgumentTypeError(
+            f"{path_str} is not a valid path.")
+    return valid_path
```
josh-hadley: At some point we should probably move stuff like this into fdkutils.py (DRY)

miguelsousa: definitely. I'll take care of it

PR #903, opened 2019-08-13
[tx] report error on short CFF2 charstring

### LINKED ISSUE #895: bug356.otf is rejected by OTS/can’t be dumped by FontTools
The font [`tests/tx_data/input/bug356.otf`](https://github.com/adobe-type-tools/afdko/blob/develop/tests/tx_data/input/bug356.otf) is rejectd by OTS and gives the follwing error with FontTools:

```
ERROR: Error in charstring 3
ERROR: Unhandled exception has occurred
  File "/fonttools/Lib/fontTools/ttx.py", line 400, in main
    process(jobs, options)
  File "/fonttools/Lib/fontTools/ttx.py", line 374, in process
    action(input, output, options)
  File "/fonttools/Lib/fontTools/misc/loggingTools.py", line 374, in wrapper
    return func(*args, **kwds)
  File "/fonttools/Lib/fontTools/ttx.py", line 271, in ttDump
    newlinestr=options.newlinestr)
  File "/fonttools/Lib/fontTools/ttLib/ttFont.py", line 227, in saveXML
    self._saveXML(writer, **kwargs)
  File "/fonttools/Lib/fontTools/ttLib/ttFont.py", line 282, in _saveXML
    self._tableToXML(tableWriter, tag, splitGlyphs=splitGlyphs)
  File "/fonttools/Lib/fontTools/ttLib/ttFont.py", line 313, in _tableToXML
    table.toXML(writer, self)
  File "/fonttools/Lib/fontTools/ttLib/tables/C_F_F_.py", line 41, in toXML
    self.cff.toXML(writer)
  File "/fonttools/Lib/fontTools/cffLib/__init__.py", line 155, in toXML
    font.toXML(xmlWriter)
  File "/fonttools/Lib/fontTools/cffLib/__init__.py", line 2382, in toXML
    self.decompileAllCharStrings()
  File "/fonttools/Lib/fontTools/cffLib/__init__.py", line 2396, in decompileAllCharStrings
    charString.decompile()
  File "/fonttools/Lib/fontTools/misc/psCharStrings.py", line 974, in decompile
    decompiler.execute(self)
  File "/fonttools/Lib/fontTools/misc/psCharStrings.py", line 290, in execute
    token, isOperator, index = charString.getToken(index)
  File "/fonttools/Lib/fontTools/misc/psCharStrings.py", line 1053, in getToken
    token, index = handler(self, b0, self.bytecode, index)
  File "/fonttools/Lib/fontTools/misc/psCharStrings.py", line 36, in read_smallInt2
    b1 = byteord(data[index])
IndexError: index out of range
```

The CharString 3 seems to be one byte too short (the last byte is 251, so there should be another byte after it). I’m not sure if the font is broken or there is a bug in both tools.

cjchapman:
resolves #895

The `bug356.otf` font originally had only one byte of a two byte integer at the end of the charstring for GID 3.

The `tx` code was treating any attempt to read past the end of a CFF2 charstring as equivalent to an `endchar` in a CFF charstring, even if it was in the middle of parsing a multi-byte operator or operand. I have changed the code to now report an error if it tries to read past the end of a CFF2 charstring while in the middle of parsing an operator or an operand. I made this fix for both the `tx -cff` and `tx -dcf` paths through the code. I also made similar fixes in some code for reading dictionaries.

I created a new folder `tests/tx_data/input/bad` for known-bad fonts, and copied the bad version of `bug356.otf` to `bug895.otf` in the `bad` folder. I then hand hex-edited the `bug356.otf` font to have a single byte integer at the end of the charstring of GID 3.

I also cleaned up several issues reported by Xcode and cppcheck. (Note that I removed some unused code as part of the cleanup.)


c/public/lib/api/tx_shared.h
```diff
@@ -526,6 +526,17 @@ struct txCtx_ {
             fatal(h, "Type 2 stack overflow");  \
     } while (0)
 
+/* Check that current charstring has enough bytes left. */
+#define CHECK_CHARSTRING_BYTES_LEFT(n)    \
```
blueshade7: We probably want to wrap these macro bodies in `do {` ...`} while (0)` in case someone attempts to write something like:
```
if (test) CHECK_DICT_BYTES_LEFT(1); else ...
```

c/public/lib/source/cffread/cffread.c
```diff
@@ -618,6 +619,11 @@ static void srcOpen(cfrCtx h, long flags, long origin, int ttcIndex) {
         if (h->stack.cnt + (n) > h->top.maxstack) fatal(h, cfrErrStackOverflow); \
     while (0)
 
+/* Check that dict has enough bytes left. */
```
blueshade7: same as above

c/public/lib/source/t2cstr/t2cstr.c
```diff
@@ -254,18 +255,20 @@ static int srcSeek(t2cCtx h, long offset) {
...
-            }                                        \
-            next = refill(h, &end);                  \
-            if (next == NULL)                        \
-                return t2cErrSrcStream;              \
-        }                                            \
+/* Check for a byte that follows the first byte of an operator or operand.
+   If we're curently empty, but more bytes are available, refill from buffer.
+   If we're completely out of bytes, throw an error.                         */
+
+#define CHECK_SUBSEQUENT_BYTE(h)                                    \
```
blueshade7: This is fine. We might want to combine the two macro definitions by passing a boolean flag to another shared macro like:
```
#define CHECK_BYTE(h, is_first)                                         \
	    do                                                              \
	        if (next == end) {                                          \
	            if (h->src.offset >= h->src.endOffset) {                \
                if (is_first && (h->aux->flags & T2C_IS_CFF2))                    \
                    goto do_endchar;                                \
...
```
And pass TRUE or FALSE from the two.

c/public/lib/api/tx_shared.h
cjchapman: Done. Thanks.

c/public/lib/source/cffread/cffread.c
cjchapman: Done. Thanks.

c/public/lib/source/t2cstr/t2cstr.c
cjchapman: I feel very strongly about having two separate macros with names that clearly state what they do, so I am going to leave this as is.

blueshade7: If so that's ok.

PR #905, opened 2019-08-14
[tx] map glyph to PUA if glyph name missing when writing svg

cjchapman:


miguelsousa:
@cjchapman Is this change already covered by an existing test?

cjchapman:
PR #908 is a follow-up to this one. It includes test cases and fixes a couple of other paths through the code.

PR #917, opened 2019-08-20
Set up CI with Azure Pipelines

miguelsousa:


miguelsousa:
The failures in **makeinstancesufo** tests are due to https://github.com/fonttools/fonttools/issues/1735

PR #920, opened 2019-08-21
[tx] added range check for Mac Roman encodings in ttread

cjchapman:
(also reduced GLYF_MAX_COMPONENT_DEPTH)


cjchapman:
@blueshade7 Thanks for the feedback and the reference point, I'll keep that in mind when I make future updates to this file.

PR #941, opened 2019-08-29
Fix #940: saveBlend() handles zero or one blend only

### LINKED ISSUE #940: mismatch between cffread.c:saveBlend() and cffwrite_dict.c:saveRealBlendOp()
In `cffread.c:saveBlend()` we’re allocating `numBlends + numRegions` values:
```
        unsigned short numBlendValues = (unsigned short)(numBlends + numRegions);
        float *blendValues = (float *)memNew(h, sizeof(abfOpEntry) * numBlendValues);
```
...but in `cffwrite_dict.c:saveRealBlendOp()` we’re trying to access `numBlends * numRegions` values:
```
    for (k = 0; k <= numRegions; k++) {
        int curIndex = (k + 1) * numBlends - 1;
        for (j = 1; j < numBlends; j++) {
            blendValues[curIndex] -= blendValues[curIndex - 1];
            curIndex--;
        }
    }
```
I thought of changing the allocation in `saveBlend()` to:
```
        unsigned short numBlendValues = (unsigned short)(numBlends * numRegions);
```
...but `saveBlend()` only writes `numRegions` values:
```
        for (i = 0; i < numRegions; i++) {
            blendValues[i + 1] = stackEntry->blend_val[i] + defaultValue;
        }
```
...so I think there’s some bigger problem here.

Attached are two fuzzed fonts produced by AFL that will cause a crash in `saveRealBlendOp()` if you build `tx` with Address Sanitizer and run it with the `-cff2` option:
[tx_cff2_crashes.zip](https://github.com/adobe-type-tools/afdko/files/3556593/tx_cff2_crashes.zip)

blueshade7:
fixes #940


c/public/lib/source/cffread/cffread.c
```diff
@@ -974,15 +974,16 @@ static void saveBlend(cfrCtx h, float *realValue, abfOpEntry *blendEntry) {
         // copy the defaul region value
         blendValues[0] = defaultValue = INDEX_REAL(0);
 
-        /* now, copy the blend abusolate values to blendArray[i].
+        /* now, copy the blend absolate values to blendArray[i].
```
cjchapman: Did you mean `absolute`?

(Otherwise these changes looks fine to me.)

blueshade7: Ah I didn't realize there are two typos in one word... will fix

PR #942, opened 2019-08-30
[comparefamily] Disable getItalicAngle function

miguelsousa:
This function depends on tx's bc mode, which has never been available since the codebase went open source 🤦‍♂️


miguelsousa:
I don't know if we care about reimplementing this functionality or not. The long term goal is to use font bakery. The short term goal is to not make a call that generates an error.

miguelsousa:
Code removed

PR #946, opened 2019-08-30
[fdkutils] Improvements & coverage

miguelsousa:


miguelsousa:
The Codacy failures are expected. The main goal of these changes is to strengthen `fdkutils.py` and raise its coverage, to enable the removal of `shell=True` with more confidence at some point in the future.

PR #955, opened 2019-09-04
[tx] fix hang in buildGIDNames

cjchapman:
(added upper bounds check on numGlyphs)


cjchapman:
After consultation with @blueshade7 and @miguelsousa, I've added a check earlier in the code, and also added a test case.

Issue #967, opened 2019-09-23
[spec] Clarify how multiple GlyphClassDef statements are handled


khaledhosny:
The feature files specification is silent on multiple `GlyphClassDef` statements. In makeotfexe, the last one overrides all the preceding ones, while in feaLib it is cumulative.

The following file:
```fea
table GDEF {
    GlyphClassDef [a], [b], [c], [d];
    GlyphClassDef [e], [f], [g], [h];
    GlyphClassDef [i], [j], [k], [l];
} GDEF;
```

Produces the following with makeotfexe:
```xml
  <GDEF>
    <Version value="0x00010000"/>
    <GlyphClassDef Format="1">
      <ClassDef glyph="i" class="1"/>
      <ClassDef glyph="j" class="2"/>
      <ClassDef glyph="k" class="3"/>
      <ClassDef glyph="l" class="4"/>
    </GlyphClassDef>
  </GDEF>
```

and the following feaLib:
```xml
  <GDEF>
    <Version value="0x00010000"/>
    <GlyphClassDef Format="1">
      <ClassDef glyph="a" class="1"/>
      <ClassDef glyph="b" class="2"/>
      <ClassDef glyph="c" class="3"/>
      <ClassDef glyph="d" class="4"/>
      <ClassDef glyph="e" class="1"/>
      <ClassDef glyph="f" class="2"/>
      <ClassDef glyph="g" class="3"/>
      <ClassDef glyph="h" class="4"/>
      <ClassDef glyph="i" class="1"/>
      <ClassDef glyph="j" class="2"/>
      <ClassDef glyph="k" class="3"/>
      <ClassDef glyph="l" class="4"/>
    </GlyphClassDef>
  </GDEF>
```

anthrotype:
I like cumulative. It is in line with the way multiple feature blocks are treated as one, allowing to merge multiple feature files with `include` statements.

readroberts:
I like cumulative as well. I vote for fixing the spec, and changing makeotf.

PR #973, opened 2019-09-29
[makeotfexe] Import fealib tests

khaledhosny:
Some tests required cosmetic changes to TTX output to pass, others had the wrong output from feaLib. The remaining xfail tests test parts of feature files spec not supported by makeotf.


anthrotype:
Wrong output from feaLib?

khaledhosny:
> Wrong output from feaLib?

https://github.com/fonttools/fonttools/pull/1733 and https://github.com/adobe-type-tools/afdko/issues/967.

PR #974, opened 2019-09-30
[tests] Increase coverage in otc2otf

miguelsousa:


tests/otc2otf_data/expected_output/RgBd.txt
```diff
@@ -0,0 +1,32 @@
+Input font: RgBd.ttc
+font 0 offset: 20/0x00000014. SourceSansPro-Regular.otf
+	BASE offset: 0x000001EC, checksum: 0x651E5DBD, length: 0x00000046
+	CFF  offset: 0x00000234, checksum: 0xF715B0B5, length: 0x0001D71D
+	DSIG offset: 0x0003C38C, checksum: 0x00000001, length: 0x00000008
```
josh-hadley: I guess this observation is more for the otf2otc side of things, but my recollection is that the OpenType specification prohibits DSIG tables in individual TTC/OTC components. TTC/OTC are supposed to have a single DSIG for the collection, none for the individual components.

miguelsousa: 🤷‍♀
Open an issue against **otc2otf** so we can discuss it, perhaps?

josh-hadley: Yes, I'll first check the spec, then see if otf2otc is actually doing something it shouldn't. I think for this specific case, where the "fake" DSIGs are used, it doesn't matter because there's no issue of invalidating the DSIG, but it could be an issue for cases where actual, valid DSIGs are involved.

readroberts:
I see that Josh  is correct, the the OpenType spec section for the DSIG table
explicitly and clearly states that only one DSIG table is allowed for the entire collection.

PR #978, opened 2019-10-01
Rewrite of otc2otf

miguelsousa:
Now it leverages FontTools as much as possible.
Option `-u` was eliminated and the formatting of the report was greatly improved.


miguelsousa:
Thanks Josh!

PR #996, opened 2019-10-14
[fontpdf] fix loop variable issue

josh-hadley:
- investigation of LGTM/Semmle-reported "unused loop variable" issue led to removal of some (but not all) uses of `eval` in fontpdf.py
- there's still a lot more in there but I'm ignoring them for now

### DIFF
--- a/python/afdko/pdflib/fontpdf.py
+++ b/python/afdko/pdflib/fontpdf.py
@@ -21,7 +21,6 @@ All coordinates are in points, in the usual PostScript model. Note,.
 however, that the coordinate system for the page puts (0,0) at the top
 right, with the positive Y axis pointing down.
 """
-
 from math import ceil
 import os
 import re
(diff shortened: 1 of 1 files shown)


miguelsousa:
Thanks @josh-hadley, this PR made me learn a new (old) thing:
https://stackoverflow.com/questions/53852422/getattr-with-two-tuples-after-it

josh-hadley:
> Thanks @josh-hadley, this PR made me learn a new (old) thing:
> https://stackoverflow.com/questions/53852422/getattr-with-two-tuples-after-it

You're welcome! In retrospect, it might have been a little clearer to do something like:
```
method = getattr(self, methodName)
method(params)
```
and
```
method = getattr(self, methodName)
method(path, pt, params)
```
...and maybe would be a little safer to check that `method` is actually callable before calling it. Maybe I'll revisit that later. But what I have now is still far safer than using `eval`.

PR #997, opened 2019-10-15
[requirements] relax pinning

### LINKED ISSUE #408: Installing may downgrade dependencies

josh-hadley:
Strict pinning with `==` in AFDKO's `requirements.txt` can result in downgrades to packages that are shared by other co-installed projects. Use of `>=` enforces only the minimum version specified and allows later versions.

Closes #408 .

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,13 +1,14 @@
 # 'lxml' is not a primary requirement of afdko but it's listed here because
 # we want to have control over the version and guarantee that the XML output
 # of our tools is stable
-lxml>=4.4.1
-booleanOperations>=0.8.2
-cu2qu[cli]>=1.6.6
-defcon[lxml,pens]>=0.6.0
-fontMath>=0.5.0
-fontTools[ufo,unicode,lxml,woff]>=4.0.1
-mutatorMath>=2.1.2
-psautohint>=2.0.0a1
-ufonormalizer>=0.3.6
-ufoProcessor>=1.0.6
+# NOTE: hard-pinning (==) here gets relaxed to >= in setup.py
+lxml==4.4.1
+booleanOperations==0.8.2
+cu2qu[cli]==1.6.6
+defcon[lxml,pens]==0.6.0
+fontMath==0.5.0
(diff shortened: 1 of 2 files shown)


anthrotype:
if you're not pinning exact versions anymore in requirements.txt, then there is no point in keeping that file. The setup.py's `install_requires` containing the same dependencies and minimum versions is sufficient.

I'm not sure if you're still using pyup bot or not, but I believe it will only update versions for requirements that have `==` version specifiers, and ignore those with `>=`.

Having a requirements.txt with fixed versions is still a good idea, because it ensures that your test runs are predictable and deterministic and do not depend on the time when you run them (e.g. it may be that some new version breaks something in your test suite, which would pass before and now longer does, simply because a new version of a dep has been uploaded to PyPI).

The setup I settled on in my projects is this: setup.py `install_requires` are open-ended and are bumped up only when needed (e.g. you require some specific symbol that has been introduced in a newer version, or a bug was fixed, etc.); whereas the requirements.txt contains exactly pinned versions that are periodically kept up-to-date with latest versions (using pyup.io bot or similar).

If you _really_ want to read the dependencies from the same source file for both `requirements.txt` and `setup.py` maybe you could resort to this hack. You keep requirements.txt as it is now, with exactly pinned versions. But then, when you read that file's content from the setup.py script, you replace all `==` with `>=` and then you'll have open-ended minimum version specifiers in `install_requires` and pinned exact versions in requirements.txt.
It's not that bad, actually.

josh-hadley:
@anthrotype I implemented the scheme you described. Please review and if this is suitable, please approve and we'll proceed to get it merged & released.

PR #998, opened 2019-10-15
[makeotfexe] confirm that nextNode is not NULL before using it

### LINKED ISSUE #993: [makeotfexe] Crash with contextual pos rule
The following (which I’m not sure is valid) crashes makeotfexe:
```fea
markClass [acute grave] <anchor 150 -10> @TOP_MARKS;                            
                                                                                
lookup mark {                                                                   
  position base [a e o u]' <anchor 250 450> mark @TOP_MARKS;                    
} mark;
```

cjchapman:
fixes #993

### DIFF
--- a/tests/makeotfexe_data/expected_output/bug993.ttx
+++ b/tests/makeotfexe_data/expected_output/bug993.ttx
@@ -1,0 +1,96 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ttFont sfntVersion="OTTO" ttLibVersion="4.0">
+
+  <GPOS>
+    <Version value="0x00010000"/>
+    <ScriptList>
+      <!-- ScriptCount=0 -->
+    </ScriptList>
+    <FeatureList>
+      <!-- FeatureCount=0 -->
+    </FeatureList>
+    <LookupList>
+      <!-- LookupCount=2 -->
+      <Lookup index="0">
+        <LookupType value="8"/>
+        <LookupFlag value="0"/>
+        <!-- SubTableCount=1 -->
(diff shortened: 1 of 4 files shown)


tests/makeotfexe_data/expected_output/bug993.ttx
```diff
@@ -0,0 +1,96 @@
...
+          <!-- BacktrackGlyphCount=0 -->
+          <!-- InputGlyphCount=1 -->
+          <InputCoverage index="0" Format="1">
+            <Glyph value="a"/>
+            <Glyph value="e"/>
+            <Glyph value="o"/>
+            <Glyph value="u"/>
+          </InputCoverage>
+          <!-- LookAheadGlyphCount=0 -->
+          <!-- PosCount=0 -->
```
anthrotype: this ChainingContextPos lookup is not doing anything. Its array of PosLookupRecords is empty.

khaledhosny: What @anthrotype said. This does not seem to be doing anything.

cjchapman: OK, that seems to be by design, as this produces it:
```
  position base [a e o u]' <anchor 250 450> mark @TOP_MARKS;                    
```
...but this (without the apostrophe) does not:
```
  position base [a e o u] <anchor 250 450> mark @TOP_MARKS;                    
```
...so I've asked @readroberts for his guidance on this.

cjchapman:
@readroberts OK. I changed both checks to throw a fatal error on NULL pointers.

PR #999, opened 2019-10-16
[fea-spec] fix example n.4 in chapter 6.h.iii

anthrotype:
the wording preceding this example refers to positioning a mark over a base glyph, so the keyword shoud be `position base` not just `position`.


anthrotype:
btw, the example is very similar to the one that crashed makeotfexe in https://github.com/adobe-type-tools/afdko/issues/993

PR #1000, opened 2019-10-16
Enhance otf2ttf for ttc, parallel, and file wildcards

msoxzw:
I mainly pick these from my [repo](https://github.com/msoxzw/fixMacFonts/blob/master/fixMacFonts/otf2ttf.py)

From my view, converting ttc font to the first font in ttc is very confusing. Hence, processing ttc font as a whole is a better way compared with splitting font, then merging it.

Moreover, calculating TrueType outline from PostScript outline is time-consuming. So, I add data parallel and file wildcards input for otf2ttf.


msoxzw:
The test error is expected, because multiprocessing module will mess up the logging module, particularly in Windows.
```
>       assert (b"WARNING: Dropping glyph names, they do not fit in 'post' "
                b"table.") in output
E       assert b"WARNING: Dropping glyph names, they do not fit in 'post' table." in b"Dropping glyph names, they do not fit in 'post' table.\r\n"
```

miguelsousa:
Looking at the failure makes me think that removing the `WARNING: ` part from the expected result should make the test pass again.

Thanks for the PR, by the way. I have a few more comments to make but I need to run some validations first.

miguelsousa:
@josh-hadley perhaps something to look into: CircleCI didn't run. I wonder if it didn't happen because the branch to be merged doesn't belong to the afdko repo.

msoxzw:
@miguelsousa 
Sorry, I do not fully understand your words. Given your attachments,  I have some confusions:

```
font.ttf    -> font#1.ttf    or font.tff    -> none

otf_ttf.ttc -> otf_ttf#1.ttc or otf_ttf.ttc -> none
              (ttf_ttf) 

ttf_otf.ttc -> ttf_otf#1.ttc or ttf_otf.ttc -> none
              (ttf_ttf)

ttf_ttf.ttc -> ttf_ttf#1.ttc or ttf_ttf.ttc -> none

```

miguelsousa:
@msoxzw if you look in https://github.com/adobe-type-tools/afdko/tree/develop/tests/otf2ttf_data/input there's only OTFs in there. The TTF and TTC files I provided in the ZIP are for adding new tests.

Here's how I expect **otf2ttf** to work with the new test input files:
* `font.ttf`, skip this file — the font is TTF already
* `ttf_ttf.ttc`, skip this file — the TTC contains 2 fonts and they're both TTF
* `otf_otf.ttc`, convert this file — the TTC contains 2 fonts and they're both OTF
* `otf_ttf.ttc`, convert this file, but skip one of its fonts — the TTC contains 2 fonts; the first font is OTF and the second is TTF
* `ttf_otf.ttc`, convert this file, but skip one of its fonts — the TTC contains 2 fonts; the first font is TTF and the second is OTF

msoxzw:
@miguelsousa
Thanks a lot. At present I have only a small question:
In `otf_ttf.ttc` or `ttf_otf.ttc`, whether or not the original TTF is also saved in the converted TTC font?

miguelsousa:
Yes, keep the original TTFs inside the TTCs.
Thank YOU for taking the time to make these enhancements.

msoxzw:
@miguelsousa
The test fonts that you provide give me much insight about OpenType.

I assume that the the results of otf_otf.ttc, otf_ttf.ttc, and ttf_otf.ttc are identical to ttf_ttf.ttc excluding checkSumAdjustment, created, and modified. However, the test error seems to detect the bugs in my codes. After a while, I think that I spot the cause of error: the 'hmtx' tables of test fonts are likely to not strictly conform to the OpenType spec, which makes some minor differences between ttf_ttf.tcc and others.

miguelsousa:
> I assume that the the results of otf_otf.ttc, otf_ttf.ttc, and ttf_otf.ttc are identical to ttf_ttf.ttc excluding checkSumAdjustment, created, and modified

Yes, I also expected that the TTX dump of **ttf_ttf.ttc** would be the same as the dump of the other TTCs after their conversion.
Looking at the failures I can see that the problem is actually in the TTCs I provided. They were made with subset fonts that didn't get all their values recalculated (e.g. `hhea.advanceWidthMax`, `maxp.maxCompositeContours`), so the TTCs wound up with inflated numbers. I'll regenerate the TTCs and provide new files soon.

python/afdko/otf2ttf.py
```diff
@@ -36,9 +40,12 @@ def glyphs_to_quadratic(
...
 
 
+@singledispatch
 def otf_to_ttf(ttFont, post_format=POST_FORMAT, **kwargs):
-    assert ttFont.sfntVersion == "OTTO"
-    assert "CFF " in ttFont
+    if ttFont.sfntVersion != "OTTO":
+        raise TTLibError("Not a OpenType font (bad sfntVersion)")
+    if "CFF " not in ttFont:
+        raise TTLibError("Not a OpenType font (no 'CFF ' table)")
```
miguelsousa: Are both of these checks required? I think only one of them is sufficient.

tests/test_utils.py
```diff
@@ -24,10 +24,14 @@ def get_bad_input_path(file_name):
...
-        temp_path = get_temp_file_path()
-        font.saveXML(temp_path, tables=tables)
-        return temp_path
+    try:
+        font = TTFont(font_path)
+    except TTLibError:
+        font = TTCollection(font_path)
+
+    temp_path = get_temp_file_path()
+    font.saveXML(temp_path, tables=tables)
```
miguelsousa: Please add `font.close()` after `font.saveXML()`.

msoxzw: I have investigated this issue before I decided to remove "with" statement. "TTCollection" class does not provide the "close" method. Therefore, adding `font.close()` is impracticable. However, due to garbage collector of Python, the font object will be automatically closed and deleted, but heavily relying on garbage collector is a bad practice. Of course, there are at least two workarounds:
1. Re-implement the close method:
```
if font.reader is not None:
   font.reader.close()
```
2. Open font file by built-in open:
```
with open(font_path, 'rb') as f:    
    try:
        font = TTFont(f)
    except TTLibError:
        font = TTCollection(f)

    temp_path = get_temp_file_path()
    font.saveXML(temp_path, tables=tables)
    return temp_path
```
Nevertheless, both methods, I think, will add  maintenance burden if the "TTCollection" class implement close method in some day.
So, how do you think?

miguelsousa:
[new_TTCs.zip](https://github.com/adobe-type-tools/afdko/files/3749166/new_TTCs.zip)

python/afdko/otf2ttf.py
msoxzw: I have ever considered one possibility what if a malicious program disguised as a font file with legal TTC header. The last important problem about font security that I have remembered is that Windows 10 change font kernel mode to user mode in order to alleviate relevant issues.

miguelsousa: OK. Doesn't hurt to leave both.

tests/test_utils.py
miguelsousa: I think the best would be to implement a `close()` method in the `TTCollection` class.
Would you mind submitting such change to https://github.com/fonttools/fonttools ?

python/afdko/otf2ttf.py
msoxzw: But I prefer to recover initial single check if any maintenance burden is excessively added. Overall, I feel somehow nervous to add these enhancements if they add unnecessary complexity.

tests/test_utils.py
msoxzw: Not at all.

python/afdko/otf2ttf.py
miguelsousa: But if you revert to the original single assert will the tool still gracefully skip TTFs, or will it halt with a traceback at the first TTF it finds?

msoxzw: Do have no influence at all except that a non-otf file with legal otf header.

miguelsousa: OK, go ahead with the solution you think it's best.

tests/test_utils.py
msoxzw: Moreover, if I submit such change and it is merged in "fonttools", this whole project will require the latest "fonttools". Hence, is it worthy to do so?

miguelsousa: Yes. The afdko always updates to the latest version of FontTools.
A lot of other projects rely on FontTools so every opportunity to make the changes there benefits everyone.

miguelsousa:
@msoxzw [code coverage on **otf2ttf** dropped quite a lot](https://codecov.io/gh/adobe-type-tools/afdko/pull/1000/src/python/afdko/otf2ttf.py?before=python/afdko/otf2ttf.py), most likely because of the parallelisation.

Do you know how to fix the reporting?

msoxzw:
@miguelsousa 
Thank you for digging out this code coverage issue from my PR. At present, this PR is waiting for the "fontTools" next release 4.0.3, and https://github.com/fonttools/fonttools/pull/1751 has been merged.

miguelsousa:
@msoxzw thanks for the FontTools contribution. We'll integrate the upcoming version shortly after it's released.

miguelsousa:
@msoxzw FontTools v4.1.0 has been integrated

msoxzw:
@miguelsousa I could not figure out why [PEP 448](https://www.python.org/dev/peps/pep-0448/) and [PEP 498](https://www.python.org/dev/peps/pep-0498/) are identified as invalid syntax in [Codacy/PR Quality Review](https://app.codacy.com/manual/adobe-type-tools/afdko/pullRequest?prid=4329700). Probably, the Python version is too old in the PR Quality Review.

miguelsousa:
@msoxzw no worries, we're actually thinking of disconnecting Codacy.

PR #1007, opened 2019-10-18
[makeotfexe] check that markClassNode is not NULL before using it

### LINKED ISSUE #1006: [makeotfexe] Crash with contextual mark to ligature
The following feature file crashes `makeotfexe`, probably similar to #993:
```fea
markClass [acute grave] <anchor 500 700> @TOP_MARKS;
lookup test {
    pos ligature c c_t' t <anchor 500 800> mark @TOP_MARKS
        ligComponent <anchor 1500 800> mark @TOP_MARKS;
} test;
```

cjchapman:
fixes #1006


khaledhosny:
I didn’t check the code, but I *think* the output is correct.

cjchapman:
@khaledhosny OK, I'd like to check the output a bit further, but I'm going to go ahead and merge this in order to fix the crash and unblock you. If the output turns out to be incorrect, I'll address it in another PR.

Issue #826, opened 2019-06-27
Remove autohint tool


miguelsousa:
Remove `autohint` tool in favor of `psautohint`.

- [x] add tests to validate that `psautohint` produces the same output as `autohint`. This will require some adjustment to `autohint` and test code:
  - [x] update `autohint` to output PFB when PFB is supplied as input (currently outputs PFA). See https://github.com/adobe-type-tools/afdko/commit/f25415a9176cbce94fa9a75ab8086da2753043a4
  - [x] add method in **test_utils.py** to use `fontTools.cffLib.specializer`, to enable comparison of equivalent Type2 Charstrings in CFF-based output (currently, output from `autohint` is optimized, while output from `psautohint` is not).
  - [x] skip comparisons where non-default-layer UFO2 is used, since `psautohint` can only write to the default layer for UFO2.
  - [x] fix bug in `autohint` where it does not delete `com.adobe.processedgyphs` layer if present when run with the `-wd` option. Fix should probably go here: https://github.com/adobe-type-tools/afdko/blob/develop/python/afdko/autohint.py#L1036
- [x] replace content in `autohint.py` that simply calls `psautohint`
- [x] add [FutureWarning](https://docs.python.org/3/library/exceptions.html#FutureWarning)s in above
- [ ] Remove `autohint.py` and related test code from AFDKO [early 2020]

PR #1011, opened 2019-10-21
[autohint] updates to prepare for psautohint

josh-hadley:
- numerous updates to autohint.py and related modules (ufotools, beztools) to prepare for eventual replacement with psautohint. This is somewhat a work in progress towards completion of #826 
- updates to tests & test data for above
- some fixes in affected modules to fix errors reported by LGTM/Semmle


python/afdko/beztools.py
```diff
@@ -1207,12 +1207,14 @@ def saveChanges(self):
 					tf.write(data)
 					tf.close()
 
-			elif  fontType == 2: # PS.
+			elif  2 <= fontType <= 3: # PS (PFA or PFB)
```
miguelsousa: would this be better?
```python
elif fontType in (2, 3):
```

tests/autohint_test.py
```diff
@@ -81,7 +80,7 @@ def test_basic_hinting(font_filename, opt):
         actual_path = get_temp_file_path()
 
     diff_mode = []
-    if tail == '.cff':
+    if tail in {'.cff', '.pfb'}:
```
miguelsousa: is there a particular reason to use a dict instead of a tuple?

josh-hadley: That's a set, not dict:
```
>>> s = {1, 2}
>>> type(s)
<class 'set'>
```

python/afdko/beztools.py
josh-hadley: Yes, that expresses the intent better, good suggestion. I updated in https://github.com/adobe-type-tools/afdko/pull/1011/commits/37a733c0a18267395d411761567573a820f4886a

PR #1021, opened 2019-11-18
[tests] add tests and data for psautohint

josh-hadley:
- _mostly_ copied from `autohint_test.py` (and autohint_data)
 - some minor (mostly formatting/inconsequential content) changes
 - removed UFO2-based tests & data
 - added `generalizeCFF` method to `test_utils.py` to enable normalized comparison of CFF/OTF
 - also updated pinning of `ufoProcessor` in requirements.txt


josh-hadley:
N.B. Codacy seems to be flagging valid Python 3.6+ f-strings as "Invalid Syntax".

PR #1029, opened 2019-11-22
[autohint][stemhist] add FutureWarnings

josh-hadley:
- add FutureWarnings to autohint.py and stemhist.py. Settled on using FutureWarning instead of DeprecationWarning so that it is displayed for all use cases (DeprecationWarning is hidden for some cases).

### DIFF
--- a/python/afdko/autohint.py
+++ b/python/afdko/autohint.py
@@ -443,7 +443,6 @@ import time
 import tempfile
 from fontTools.ttLib import TTFont, getTableModule
 import plistlib
-import warnings
 from afdko import fdkutils, ufotools
 from afdko.beztools import *
 import traceback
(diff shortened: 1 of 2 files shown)


python/afdko/autohint.py
```diff
@@ -448,12 +448,18 @@
...
+import warnings
 
 try:
     from psautohint import AUTOHINTEXE
 except ImportError:
     AUTOHINTEXE = "autohintexe"
 
+warnings.warn(
+	"autohint has been deprecated and will be removed from AFDKO soon. "
+	"Please update your code.",
```
miguelsousa: What about `Please update your code to use psautohint.`?

josh-hadley: Good idea, done in https://github.com/adobe-type-tools/afdko/pull/1029/commits/6f57a3b5c43863c2c748f34362d5fbe6344e04bd. Also fixed that duplicate import that LGTM/Semmle picked up.

PR #1062, opened 2019-12-13
fix several instances of "declaration hides parameter"

cjchapman:


c/makeotf/makeotf_lib/source/cffread/cffread.c
```diff
@@ -1977,11 +1977,11 @@ static void DICTRead(cffCtx h, int length, Offset offset, int enable) {
                         h->stack.cnt = 0;
                         if (!(h->flags & WV_SET)) {
                             /* Convert UDV to WV if WV not already set */
-                            unsigned length;
+                            unsigned string_length;
```
josh-hadley: I assume this will default to `int` but does it make sense to be explicit here?

cjchapman: An `unsigned` by itself means the same thing as an `unsigned int`. This is defined in the C standard, and it is used in hundreds of places in the AFDKO, so I'm reluctant to change it.

josh-hadley: 👍

PR #1094, opened 2020-02-19
[waterfallplot] fix crash

### LINKED ISSUE #1092: [waterfallplot] crashes if requested waterfall range overflows
`waterfallplot` doesn’t like the requested waterfall range to exceed the physical page size.
In other words: the waterfall doesn’t cascade. 

`waterfallplot -wfr 60-70 font.pfa`
```
Proofing font font.pfa. Start time: Tue Feb 18 14:02:22 2020.
  File "/Users/fgriessh/.pyenv/versions/3.7.4/bin/waterfallplot", line 8, in <module>
    sys.exit(waterfallplot())
  File "/Users/fgriessh/.pyenv/versions/3.7.4/lib/python3.7/site-packages/afdko/proofpdf.py", line 945, in waterfallplot
    main()
  File "/Users/fgriessh/.pyenv/versions/3.7.4/lib/python3.7/site-packages/afdko/proofpdf.py", line 972, in main
    proofMakePDF(params.rt_fileList, params, txPath)
  File "/Users/fgriessh/.pyenv/versions/3.7.4/lib/python3.7/site-packages/afdko/proofpdf.py", line 893, in proofMakePDF
    pdfFilePath = makePDF(pdfFont, params, doProgressBar)
  File "/Users/fgriessh/.pyenv/versions/3.7.4/lib/python3.7/site-packages/afdko/pdflib/fontpdf.py", line 1670, in makePDF
    makeWaterfallPDF(params, pdfFont, doProgressBar)
  File "/Users/fgriessh/.pyenv/versions/3.7.4/lib/python3.7/site-packages/afdko/pdflib/fontpdf.py", line 1992, in makeWaterfallPDF
    numPages = int(ceil(float(numWaterFalls) / numWaterfallsOnPage))
ZeroDivisionError: float division by zero
```

Test font attached:
[font.pfa.zip](https://github.com/adobe-type-tools/afdko/files/4219321/font.pfa.zip)

josh-hadley:
update pdflib/fontpdf.py to address crash
closes #1092

### DIFF
--- a/python/afdko/pdflib/fontpdf.py
+++ b/python/afdko/pdflib/fontpdf.py
@@ -1993,7 +1993,7 @@ def makeWaterfallPDF(params, pdfFont, doProgressBar):
 	if numWaterfallsOnPage > 0:
 		numPages = int(ceil(float(numWaterFalls) / numWaterfallsOnPage))
 	else:
-		numPages = int(ceil(float(numWaterFalls)*pageHeight / waterfallHeight))
+		numPages = int(ceil(float(numWaterFalls) * waterfallHeight / pageHeight))
 
 	if numPages == 0:
 		numPages = 1
--- a/python/afdko/pdflib/fontpdf.py
+++ b/python/afdko/pdflib/fontpdf.py
@@ -1989,7 +1989,12 @@ def makeWaterfallPDF(params, pdfFont, doProgressBar):
 
 	numWaterfallsOnPage = int(pageHeight / waterfallHeight)
 	numWaterFalls = len(glyphLists)
-	numPages = int(ceil(float(numWaterFalls) / numWaterfallsOnPage))
+
+	if numWaterfallsOnPage > 0:
+		numPages = int(ceil(float(numWaterFalls) / numWaterfallsOnPage))
+	else:
+		numPages = int(ceil(float(numWaterFalls)*pageHeight / waterfallHeight))
+
 	if numPages == 0:
 		numPages = 1
 	doTitle(rt_canvas, pdfFont, params, numGlyphs, numPages)


python/afdko/pdflib/fontpdf.py
```diff
@@ -1989,7 +1989,12 @@ def makeWaterfallPDF(params, pdfFont, doProgressBar):
 
 	numWaterfallsOnPage = int(pageHeight / waterfallHeight)
 	numWaterFalls = len(glyphLists)
-	numPages = int(ceil(float(numWaterFalls) / numWaterfallsOnPage))
+
+	if numWaterfallsOnPage > 0:
+		numPages = int(ceil(float(numWaterFalls) / numWaterfallsOnPage))
+	else:
+		numPages = int(ceil(float(numWaterFalls)*pageHeight / waterfallHeight))
```
cjchapman: I think the code at line 1996 should be:
```c
numPages = int(ceil(float(numWaterFalls) * waterfallHeight / pageHeight))
```
to be more like the calculation in line 1994.

Alternatively, perhaps you could replace lines 1993-1996 with:
```c
if numWaterfallsOnPage > 0:
	numWaterfallsOnPage = 1
numPages = int(ceil(float(numWaterFalls) / numWaterfallsOnPage))
```
similar to what is done in lines 1998-1999.

josh-hadley: I tried both of these suggestions locally using the example font and command from #1092 , and the only difference that I'm able to discern is the total page count displayed in the header.

In my original commit (using code suggested by @readroberts ), the total page count is shown as 16 (which is incorrect; the actual number of pages is 28).

Using the first method suggested above (change line 1996), the total page count is shown as 26, which is closer, but still wrong. For the alternative method above, the page count is 20 (NOTE: I tried to attach the PDF results of both of these, but GitHub is not letting me right now).

I want to be clear here that the _actual number of pages generated_ for all three of these is 28, and the waterfall content on the pages is identical, apart from the "## of ##" page indicator and timestamp in the headers.

PR #1098, opened 2020-03-03
[ufotools] remove unused code

josh-hadley:
Removed unused methods, imports, and classes from ufotools.py module which are no longer used since autohint has been deprecated. This has a side-effect of improving test code coverage.


python/afdko/ufotools.py
```diff
@@ -1254,135 +825,6 @@ def parsePList(filePath, dictKey=None):
...
-    if (firstX == lastX) and (firstY == lastY):
-        del outlineItem[-1]
-        firstPointElement.set("type", lastOp)
-    else:
-        # we have an implied final line to. All we need to do is convert
-        # the inital moveto to a lineto.
-        firstPointElement.set("type", "line")
-
-
 bezToUFOPoint = {
```
miguelsousa: `bezToUFOPoint` dict can be removed as well, I think.

josh-hadley: done (and more) in https://github.com/adobe-type-tools/afdko/pull/1098/commits/058525a6076ffd5eb769598a6e613a7667a1ed20

PR #1126, opened 2020-04-29
Add option to use varLib instead of MutatorMath in makeinstancesufo

punchcutter:
Add makeinstancesufo test for MutatorMath 3.0.1 and varLib


punchcutter:
This test should randomly fail because of the way ufoProcessor sets the public.glyphOrder when instantiating.

miguelsousa:
@punchcutter Add `<lib copy="1"/>` to the default master in the designspace file. That should make all instances have the same glyph order as the default master.

PR #1127, opened 2020-04-29
STAT table support

### LINKED ISSUE #167: [spec] Malformed example in section 6.b.i
In [section 6.b.i](http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html#6.b.i), the OpenType Feature File Syntax specification has the following example:

```
  position T <-60> a <-40 0 -40 0>;
```

This is invalid syntax; makeotf correctly rejects it. Instead, the example should say:

```
  position T -60 a <-40 0 -40 0>;
```

### LINKED ISSUE #176: Implement support for syntax for STAT table
Could the AFDKO feature syntax support a `table STAT` statement? See @moyogo’s [proposal](https://github.com/LettError/designSpaceDocument/issues/8#issuecomment-293540107) to encode STAT in [designspace documents](https://github.com/LettError/designSpaceDocument/blob/master/README.md), but @LettError didn’t like it (because he thinks it it adds too much cruft to the designspace format).

khaledhosny:
Fixes #176.

Almost done, except for name id sharing (still trying to figure out how to reliably do that). Opening to see if CI is happy with what I’ve got so far.


khaledhosny:
The LGTM alerts are in ANTLR generated code, is there away to silence these?

josh-hadley:
> The LGTM alerts are in ANTLR generated code, is there away to silence these?

There is a way to do it: https://lgtm.com/help/lgtm/customizing-file-classification

But, worth noting: the LGTM checks are informative only. They won't block the PR. So feel free to just ignore.

cjchapman:
@khaledhosny I just made this change on `develop` which should help for future PRs:
https://github.com/adobe-type-tools/afdko/blob/584fab7d467438326c6081093d72c84d96996c01/.lgtm.yml#L13-L16

khaledhosny:
> But, worth noting: the LGTM checks are informative only. They won't block the PR. So feel free to just ignore.

It was a bit annoying since it kept adding new comments each time I push new changes.

> @khaledhosny I just made this change on `develop` which should help for future PRs:

Thanks. I rebased my branch it helps indeed.

khaledhosny:
I’m trying to implement name ID sharing, but all ideas I can come up with seem unnecessarily complicated:
* It has to be deferred to table compilation time otherwise defining `STAT` before `name` will give different results.
* This means lots of changes to how the `name` table is built since the name ID will be known only when the table is about to be compiled not when the name is added, unlike the case now.
* The criteria of when to share and when not gives me headache, with all the possible combinations of platform, language, and encoding IDs.

Also the value of such sharing might not be that big:
* There is not much file saving, the actual strings are always shared even if used for different name IDs, so we are only saving the 12 bytes of the name record.
* There is already `ElidedFallbackNameID`, and other user name (e.g. for feature parameters don’t  do such sharing).

Sorry for always coming back to this question :smile: but do we really want to to do this :smile: Last time to ask I promise.

josh-hadley:
> I’m trying to implement name ID sharing, but all ideas I can come up with seem unnecessarily complicated:
> [...]
> Sorry for always coming back to this question 😄 but do we really want to to do this 😄 Last time to ask I promise.

OK, let's proceed without nameRecord sharing for now and at least get the basic implementation integrated. We can defer the sharing feature until later or possibly devise a post-process action to handle it.

Thanks for your effort on this, @khaledhosny !

khaledhosny:
All, good then. This should be more or less done, though I’d like to see #1128 before merging to check if the test coverage for the new code is good or not. The code is otherwise ready for review, and the documentation should follow once we are happy with the code.

josh-hadley:
>  I’d like to see #1128 before merging to check if the test coverage for the new code is good or not.

I will try to work that out soon (via GHA, using a similar technique as I did in psautohint).

cjchapman:
@khaledhosny Sorry, looks like I made a mistake in the path to featgram.c in the .lgtm.yml. I think I've fixed it on `develop` in [this commit](https://github.com/adobe-type-tools/afdko/commit/fbde14b27292fcb0c775e5f257f8491760f3a310).

khaledhosny:
I think the code is ready for review and the test coverage is good.

punchcutter:
@khaledhosny Did you do update the spec as well? I don't see that update in here, but the tests reference the new sections of the spec.

khaledhosny:
I was waiting until the code is finalized, but I guess it is time to do it.

punchcutter:
Ok, no problem either way. Just wanted to double check.

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
...
+There can be more than one `location` statement when this format is used (in
+this case the axis value will be `STAT` format 4 `AxisValue`, and format 1
+otherwise).
+
+##### location format B
+```fea
+location <axisTag> <nominalValue> <rangeMinValue> - <rangeMaxValue>;
+```
+
+Format for `axisTag` and the other values as above, and there must be space
```
punchcutter, line 3542: Should this say "there must be a space before and after the `-` to distinguish it from a possible minus sign for `<rangeMaxValue>`" ?

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
...
+    name <name spec>;+
+};
+```
+
+The syntax for the individual name string entries is similar to that of the
+name table `nameID` entries (see see §[9.e](#9.e)) - the only difference is
+that the introductory keyword is `name`, and the name ID value is omitted,
+since the nameID value is auto-generated by the feature compiler.
+
+Alternatively if the name string matches one of the pre-existing names in the
```
punchcutter, line 3481: Add comma after Alternatively -> `Alternatively,`

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
...
+    name <name spec>;+
+};+
+```
+
+`axisTag` is a four-letter tag. `axisOrdering` is a signed number.
+
+The syntax for individual name string entries is the same as above.
+
+#### Axis values
+There can one or more axis values specified in the `STAT` table.
```
punchcutter, line 3503: There can _be_ one or more axis values...

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
...
+`flag` is optional, and defaults to 0 when omitted. Possible flags are
+`ElidableAxisValueName` and `OlderSiblingFontAttribute`, and more than one flag
+can be specified, e.g:
+
+
+```fea
+flag ElidableAxisValueName OlderSiblingFontAttribute;
+```
+
+There are several formats for `location` statement:
```
punchcutter, line 3522: I know it's not totally consistent, but other parts of the spec might say it like 
The `location` statement takes several formats:
Or explicitly say how many.

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
...
+
+##### location format A
+```fea
+location <axisTag> <value>;+
+```
+
+`axisTag` is a 4-letter tag and must correspond to one of the defined design
+axes in the table. `value` is signed number and can be specified using decimal,
```
punchcutter, line 3530: `value` is _a_ signed number

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
...
+   AxisValue {
+      location wght 400;
+      name "Regular";
+      name 3 1 0x411 "\5B9A\671F\7684";
+      flag ElidableAxisValueName;
+   };
+
+   # format 2
+   AxisValue {
+      location wght 400 300- 500;
```
punchcutter, line 3580: Should this hyphen have a space before it?

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
...
+      location ital 0 1;
+      name "Regular";
+      flag ElidableAxisValueName;
+   };
+
+   # format 4
+   AxisValue {
+      location wght 500;
+      location ital 1;
+      name "Mediumitalic";
```
punchcutter, line 3603: Totally doesn't matter at all, but it bothers me that the **i** is lowercase. Could it be `MediumItalic`?

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
 horizontal advance of their corresponding target (upright) glyphs. These values
 will also be overridden by `VertAdvanceY` values.
 
+<a name="9.i"></a>
+### 9.i. STAT table
+STAT table is organized as followas:
```
punchcutter, line 3456: I'd say "The STAT table"
"followas" should be "follows"

docs/OpenTypeFeatureFileSpecification.md
khaledhosny: The grammar does not require space after the number, but requires space before the hyphen otherwise it is would be considered a minus sign.

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
...
+defined using full name entries or an existing name ID.
+
+```fea
+ElidedFallbackName {
+    name <name spec>;+
+};
+```
+
+The syntax for the individual name string entries is similar to that of the
+name table `nameID` entries (see see §[9.e](#9.e)) - the only difference is
```
punchcutter, line 3477: Oops, didn't catch the double `see see`

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3442,6 +3451,161 @@ the `vrt2` feature. The default vertical advance for these glyphs is the
...
+AxisValue {
+    flag <FLAG>+ ;
+    name <name string>;+
+    location ...;+
+};+
+```
+
+`flag` is optional, and defaults to 0 when omitted. Possible flags are
+`ElidableAxisValueName` and `OlderSiblingFontAttribute`, and more than one flag
+can be specified, e.g:
```
punchcutter, line 3515: `e.g.,` or maybe just `for example` is fine

docs/OpenTypeFeatureFileSpecification.md
khaledhosny: The typo is an an old paragraph where the text was copied and slightly edited. I’m fixing three 3 occurrences of it.

punchcutter: Great, thanks.

PR #1133, opened 2020-05-17
Allow UTF-8 input for name entries

### LINKED ISSUE #165: [spec] Specify Feature File encoding as UTF-8
The current [Feature File Syntax](http://www.adobe.com/devnet/opentype/afdko/topic_feature_file_syntax.html) does not say what encoding a feature file should have, besides restricting string literals to ASCII.

Proposal: Change the spec to require that feature files use UTF-8 encoding; allow for arbitrary Unicode string literals; and change section 9.e to allow for Unicode strings in `nameid` statements.

This would preserve backwards compatibility as long as the current escaping mechanism is left unchanged. For example, the following two blocks would produce the exact same output. The former is from the current spec, and the escapes would still continue to be based on platform encodings even after the suggested specification change. However, font designers could also use the new syntax below, and it would be up to the compiler to convert this whatever is needed for the OpenType table.

```
  table name {
     nameid 9 "Joachim M\00fcller-Lanc\00e9";    # Windows (Unicode)
     nameid 9 1 "Joachim M\9fller-Lanc\8e";      # Macintosh (Mac Roman)
  } name;

  table name {
     nameid 9 "Joachim Müller-Lancé";    # Windows (Unicode)
     nameid 9 1 "Joachim Müller-Lancé";  # Macintosh (Mac Roman)
  } name;
```

In https://github.com/fonttools/fonttools/issues/780#issuecomment-279654676, @twardoch suggested an extension to feaLib. Personally I like Adam’s idea (minus the complexity of arbitrary encodings; simply requiring UTF-8 would be easier). But I’d prefer to not deviate from the published feature file spec, hence filing this issue here.

khaledhosny:
Fixes #165.

UTF-8 comments seem to work fine, I’m not aware for other places where UTF-8 input is rejected.


khaledhosny:
Several failures seem to be caused by https://github.com/fonttools/fonttools/issues/1964 which is unrelated to this PR, and CircleCI linux-python3 is failing to lint `setup.py` which is also unrelated to this PR.

The most unfortunate PR I ever made, catching two unrelated CI failures :smile:.

josh-hadley:
> @khaledhosny can you please bump `MAKEOTF_VERSION` and `HOT_VERSION`?

Also `__version__` attribute (version & date) in makeotf.py; even though the Python code didn't change, just so we reflect a change in the overall makeotf toolchain.

josh-hadley:
Something very strange is going on with dependencies. The GHA Test suites, which explicitly `pip install -r requirements.txt`, work fine. But GHA Coverage and other CIs fail with bizarre failures on code that hasn't been touched in this PR. I experienced something like this locally earlier. Need to investigate, something in the packaging/setup scheme must have changed.

punchcutter:
Tests are failing because fontTools changed SFNTReader and removed a required import of SimpleNamespace.

anthrotype:
It should be fixed in fonttools 4.10.2, apologies for inconvenience

josh-hadley:
@khaledhosny this is indeed an unfortunate PR 😸 Looks like the fontTools 4.10.2 update has [another change](https://github.com/fonttools/fonttools/commit/db26cf804e3380b5e9caed3a33d640b8e8a54835) that causes one of our tests to fail still (unrelated to your code). I am going to get that fixed so this PR is clean (tests passing) and get this merged ASAP before something else happens!

anthrotype:
> Looks like the fontTools 4.10.2 update has another change that causes one of our tests to fail still 

As for the above mentioned change with findMultilingualName this time it should be a feature rather than a bug.

anthrotype:
For the records, see https://github.com/fonttools/fonttools/pull/1921#issuecomment-624699000

josh-hadley:
> As for the above mentioned change with findMultilingualName this time it should be a feature rather than a bug.

Understood. That's what I got from Just's log message, and is why I'm updating our test code (actually it looks like we have the same test case or at least the same name strings, because the changes to the fontTools test case are exactly the failures we have currently).

josh-hadley:
Thanks for your work on this @khaledhosny and sorry it was such a pain to get finished! Smoother ride next time 🤞

khaledhosny:
Thanks @josh-hadley. It wasn’t that painful, it was actually a bit funny :smile:

PR #1136, opened 2020-05-22
[spot] increase workstr buffer size and use snprintf to prevent overflow

cjchapman:
(this fixes a crash that would occur with very long file paths)


josh-hadley:
The GHA failures are expected and OK as this PR doesn't have the `rename()` fix for `sfntedit` that I have in the works.

Issue #1125, opened 2020-04-29
fontsetplot/proofpdf -g broken when using multiple fonts


GTMxCode:
Hey, been trying to make use of and understand these functions and how they can be of use. After finding out that waterfallplot doesnt work with ttfs and after a lot of trudging through the depths to figure out why and what I could do to maybe make it work with them I gave up and figured what was more necessary was comparison across fonts for now. So I set out to use fontsetplot but I ran into this issue. 

The fonts im trying to compare have upwards of 7000 glyphs and I dont need them all displayed, what I wanted was just the standard a-Z, numbers, punctuation and maybe some other chars. I started simply by trying to output a-Z,zero-nine
`fontsetplot -g a-z,A-Z,zero-nine font1.ttf font2.ttf ....`

Which generates something like:

> Collecting font data from:
>         font.ttf.
>         font2ttf.
> Writing PDF pages...
> Traceback (most recent call last):
>   File "/home/gtm/.local/bin/fontsetplot", line 8, in <module>
>     sys.exit(fontsetplot())
>   File "/home/gtm/.local/pipx/venvs/afdko/lib/python3.6/site-packages/afdko/proofpdf.py", line 935, in fontsetplot
>     main()
>   File "/home/gtm/.local/pipx/venvs/afdko/lib/python3.6/site-packages/afdko/proofpdf.py", line 972, in main
>     proofMakePDF(params.rt_fileList, params, txPath)
>   File "/home/gtm/.local/pipx/venvs/afdko/lib/python3.6/site-packages/afdko/proofpdf.py", line 840, in proofMakePDF
>     pdfFilePath = makeFontSetPDF(pdfFontList, params, doProgressBar)
>   File "/home/gtm/.local/pipx/venvs/afdko/lib/python3.6/site-packages/afdko/pdflib/fontpdf.py", line 2156, in makeFontSetPDF
>     pdfGlyph = pdfFont.getGlyph(glyphName)
>   File "/home/gtm/.local/pipx/venvs/afdko/lib/python3.6/site-packages/afdko/pdflib/fontpdf.py", line 464, in getGlyph
>     return self.clientGetGlyph(glyphName)
>   File "/home/gtm/.local/pipx/venvs/afdko/lib/python3.6/site-packages/afdko/pdflib/ttfpdf.py", line 60, in clientGetGlyph
>     return txPDFGlyph(self, glyphName)
[...]

PR #1148, opened 2020-05-30
fontsetplot and ttfdecomponentizer fixes

miguelsousa:
Fixes problems raised in #1125.


miguelsousa:
@josh-hadley good catch! Option 1 seemed the best to me, so I went with that.

Issue #992, opened 2019-10-12
[makeotfexe] ignore rsub is resolved to ignore sub


khaledhosny:
Not sure if this intentional or not, but `ignore rsub` produces the same lookup as `ignore sub`. The following two lookups:
```fea
lookup test {                                                                   
 ignore rsub A [B];                                                             
} test;
```
```fea
lookup test {                                                                   
 ignore sub A [B];                                                             
} test;
```

Produce:
```xml
  <GSUB>                                                                        
    <Version value="0x00010000"/>                                               
    <ScriptList>                                                                
      <!-- ScriptCount=0 -->                                                    
    </ScriptList>                                                               
    <FeatureList>                                                               
      <!-- FeatureCount=0 -->                                                   
    </FeatureList>                                                              
    <LookupList>                                                                
      <!-- LookupCount=1 -->                                                    
      <Lookup index="0">                                                        
        <LookupType value="6"/>                                                 
        <LookupFlag value="0"/>                                                 
        <!-- SubTableCount=1 -->                                                
        <ChainContextSubst index="0" Format="3">                                
          <!-- BacktrackGlyphCount=0 -->                                        
          <!-- InputGlyphCount=1 -->                                            
          <InputCoverage index="0" Format="1">                                  
            <Glyph value="A"/>                                                  
          </InputCoverage>                                                      
          <!-- LookAheadGlyphCount=1 -->                                        
          <LookAheadCoverage index="0" Format="1">                              
            <Glyph value="B"/>                                                  
          </LookAheadCoverage>                                                  
          <!-- SubstCount=0 -->                                                 
        </ChainContextSubst>                                                    
      </Lookup>                                                                 
    </LookupList>                                                               
  </GSUB>              
```

readroberts:
Definitely a bug. The rsub should show a different lookup type.

NorbertLindenberg:
In addition, when `ignore rsub` is used together with another `rsub` rule, this leads to weird behavior that’s hard to understand:

* If the two rules are inside a `lookup` block, makeotf reports a bug `Error: "Lookup type different from previous rules in this lookup block" in Feature mark in line: xx`. As usual with makeotf, `xx` is far far from the location of the code that it misunderstands.

* If the two rules are outside a `lookup` block, makeotf doesn’t complain, but quietly packages them in two separate lookups, thus defeating the purpose of the `ignore` rule.

khaledhosny:
[...]

Issue #1005, opened 2019-10-18
[makeotfexe] NULL in substitutions?


khaledhosny:
The following lines:
https://github.com/adobe-type-tools/afdko/blob/414ef3b23707b187e22630198dbfe5032d10b6fa/c/makeotf/makeotf_lib/build/hotpccts/featgram.g#L887-L896

and 
https://github.com/adobe-type-tools/afdko/blob/414ef3b23707b187e22630198dbfe5032d10b6fa/c/makeotf/makeotf_lib/build/hotpccts/featgram.g#L900-L912

Suggest that the following syntax should be supported:
```fea
sub A by NULL;
sub A from NULL;
rsub A by NULL;
```

But in the three cases `makeotfexe` fails with:
```
contextual substitution clause must have a replacement rule or direct lookup reference.
```

So either the grammar rules are dead code or this is an undocumented feature that is currently broken.

PR #1158, opened 2020-06-09
[makeotf] Use correct lookup type for `ignore rsub`

khaledhosny:
The code was unconditionally overriding the lookup type to become Chaining Contextual Substitution, instead do it only if it is not Reverse Chaining Contextual Substitution.

Fixes https://github.com/adobe-type-tools/afdko/issues/992


khaledhosny:
While doing some cleanup to the code, I also fixed https://github.com/adobe-type-tools/afdko/issues/1005 by dropping the nonworking and undocumented special support for `NULL` substitution.

tests/makeotfexe_data/input/spec/5fii-4.fea
```diff
@@ -0,0 +1,5 @@
+# Test comma-separated ingore sub statements.
```
punchcutter, line 1: typo _ignore_

tests/makeotfexe_data/input/spec/5h-6.fea
```diff
@@ -0,0 +1,5 @@
+# Test comma-separated ingore rsub statements.
```
punchcutter, line 1: typo _ignore_

khaledhosny:
I didn’t actually do any testing, I just verified that the generated GSUB table look as expected (a Reverse Substitution Lookup with no substitutions. The semantics of the text file may be nonsense, I’ll need to heck this and see what should be the expected behavior.

khaledhosny:
Here is my finding so far, and `ReverseChainSingleSubstFormat1` subtable with empty `substituteGlyphIDs` will be ignore by Uniscribe and Core Text, but HarfBuzz will use GID 0 (`.notdef`) for substitute glyph (looks like a bug), so I guess empty substitutes is an unsupported case (unlike chaining substitution subtables, reverse chaining does not reference other lookups but includes the substitutes directly which might explain the difference).

This might be the reason `ignore rsub` was treated as if it was `ignore sub`, but doing this makes it useless as it would do nothing (as we end with two lookups, not one lookup with two subtables).

I think our options either:
1. disallow `ignore rsub` and make it an error, or
2. make `ignore rsub A' B;` translate to `rsub A' B by A;` (i.e. replace A by itself), which seems to give the desired effect in HarfBuzz and Uniscribe but ignored by Core Text, or
3. Keep things as they are now (broken) since ~no one is complaining.

khaledhosny:
Since `ignore rsub` apparently never worked, and making it work is a bit non-trivial and might be surprising since it is not the same as `ignore sub`, I’m leaning towards option 1). The same effect can be achieved by `rsub A' by A;` which is more explicit and if it cause any surprises they should be more visible to the user.

punchcutter:
I wonder how many people are using rsub in the first place. @khaledhosny are you using it?

khaledhosny:
I’m using `rsub` in one font, yes, and it is working fine. `ignore rsub` is the broken one, and I think it was an untested feature under the mistaken believe that reverse chaining substitution is similar in structure to chaining substitution.

My suggestion would to drop `ignore rsub` from the spec and the code since it never worked and can’t be made to work.

punchcutter:
Yes, I know we are talking about `ignore rsub`, but it goes together with `rsub` so I'm wondering how common it is. Do you have an example of how you are using it that it's working fine?

khaledhosny:
`rsub` is rather rarely used, I’ve only used it once but the project is not publicly available right now. But here is the part using `rsub`. 

```fea
lookup BehAlefOrLam {
  sub [behDotless-ar.init behDotless-ar.medi]'
      [lam-ar.medi]
   by [behDotless-ar.init.prep behDotless-ar.medi.prep];
} BehAlefOrLam;

lookup BehAlefOrLamR {
 rsub [behDotless-ar.init behDotless-ar.medi]'
      [behDotless-ar.init.prep behDotless-ar.medi.prep]
   by [behDotless-ar.init.prep behDotless-ar.medi.prep];
} BehAlefOrLamR;
```
Which basically changes medial `beh` to a unslanted alternate when followed by `lam` (the `sub`), and changes any `beh` before the alternate one to a similar form (the `rsub` part) for any arbitrary sequence of `beh`s (since `rsub` works on the glyph sequence from last to first).

image

image

josh-hadley:
Closing, per discussion. See #1225.

PR #1165, opened 2020-06-12
[tx] don't write FontMatrix in CFF2 FontDict

josh-hadley:
- update cffwrite_dict.c so that it does not write FontMatrix in cfwDictFillFont in the CFF2 case
- added tests for above (CFF and CFF2 output)
- updated cffwrite version to 1.0.56


anthrotype:
Thanks!

Issue #1145, opened 2020-05-27
[makeinstancesufo] should surface more info when failing


frankrolf:
I just encountered my good friend #790 again, in the middle of a `makeinstancesufo` session.
I have no idea if there would be a way to make this happen, but it would be useful to know what exactly makes the process fail, right away.

Here is my journey (please note that the whole makeinstancesufo process takes several minutes):

1 . In non-verbose mode, the only output is `ERROR:afdko.makeinstancesufo:index must represent an on curve point`, without even specifying what instance might contain the issue.

2. Using the `-v`-mode in a second complete run, I can at least find out which instance we are talking about.
```
INFO:_psautohint:f.ligalong: Removed hints from short element at -6 -232.
INFO:_psautohint:tonos.cap: Removed hints from short element at 76 449.
INFO:_psautohint:turkicdsccmb: Removed conflicting hints at 32 -33.
INFO:psautohint.autohint:Done with font /Users/fgriessh/work/source-serif-pro/Italic/Instances/CaptionBlackIt/font.ufo. End time: Wed May 27 14:35:03 2020.
INFO:afdko.makeinstancesufo:Doing overlap removal with checkoutlinesufo on /Users/fgriessh/work/source-serif-pro/Italic/Masters/../Instances/ExtraLightIt/font.ufo ...
ERROR:afdko.makeinstancesufo:index must represent an on curve point
```

3. The `-vv` mode doesn’t add anything useful to this:

```
makeinstancesufo -i 6 -vv -d Italic/Masters/SourceSerifPro-Italic.designspace
INFO:afdko.makeinstancesufo:Building 1 instance...
INFO:afdko.makeinstancesufo:Built 1 instances.
INFO:afdko.makeinstancesufo:Applying UFO normalization...
DEBUG:ufonormalizer:Normalizing "glyphs/contents.plist".
DEBUG:ufonormalizer:Normalizing "glyphs/layerinfo.plist".
DEBUG:ufonormalizer:Normalizing "metainfo.plist".
DEBUG:ufonormalizer:Normalizing "fontinfo.plist".
DEBUG:ufonormalizer:Normalizing "groups.plist".
DEBUG:ufonormalizer:Normalizing "kerning.plist".
DEBUG:ufonormalizer:Normalizing "layercontents.plist".
DEBUG:ufonormalizer:Normalizing "lib.plist".
INFO:afdko.makeinstancesufo:Applying post-processing...
[...]

PR #1170, opened 2020-06-15
[checkoutlinesufo, ufotools] improve overlap removal, error messages

josh-hadley:
- modify check for contour starts (base contour indices on original glyph, not fixed glyph)
- update `thresholdAttrGlyph` method: don't try to "fix" smooth points, use threshold of 0.5 **N.B. this change causes changes to test data** w/r/t smoothed points
- update to use `ufoVersionFormatTuple` consistently throughout module
- improve messaging for KeyError and failed assertion in overlap removal/contour renumbering scheme


josh-hadley:
@frankrolf @punchcutter I would like to have you both review this if possible -- this addresses several issues on https://github.com/adobe-type-tools/afdko/issues/1145 that Frank has been following. I have another related branch that will also apply to that effort.

PR #1172, opened 2020-06-21
[makeotf] Check for PostScript name in FontMenuNameDB file

### LINKED ISSUE #1171: FontMenuNameDB behavior can be a bit confusing
I remember the `FontMenuNameDB` file used to be an integral file required to build an OTF – even for just building a single style. `makeotf` would fail without a `FontMenuNameDB` (and without a `GlyphOrderAndAliasDB`).

A while ago this changed, so `makeotf` can be called without much of that collateral – font naming will be deduced from the PS name within the UFO.

Today I noticed that the presence of a `FontMenuNameDB` which has _unrelated_ names in it can make `makeotf` fail.
See attached projects for an example.

- Both projects contain a `font.ufo` file with PS name `abc-def`
- When adding this FontMenuNameDB (`does not work.zip`) project, `makeotf` will fail:

```
[Unrelated-Fontname]
	f=Unrelated
	s=Fontname
```

This behavior could be a deliberate decision, or an oversight – I just wanted to put it up for discussion.

[does work.zip](https://github.com/adobe-type-tools/afdko/files/4799818/does.work.zip)
[does not work.zip](https://github.com/adobe-type-tools/afdko/files/4799819/does.not.work.zip)

miguelsousa:
Fixes #1171


frankrolf:
This did not work as expected. I did the following:

```
# install and activate venv
python -m venv fmndb_update
source fmndb_update/bin/activate

# install and verify branch
pip install git+https://github.com/adobe-type-tools/afdko.git@ms-fmndb-bug1171
which makeotf  # -> /Users/fgriessh/Desktop/fmndb_update/bin/makeotf

# cd to directory where a UFO file is
makeotf -f Exercise3_0.ufo

# output
makeotf [Warning] Could not find default features file. Font will be built without any layout features.
makeotfexe [WARNING] not in FontMenuNameDB [Exercise3-Master0]
makeotfexe [FATAL] <Exercise3-Master0> I can't find a Family name for this font !
Error executing command 'makeotfexe -f /var/folders/7z/vl3dc3kd07bfz1q_5ybyk5wm0000gn/T/tmpa8w09wt5 -o /Users/fgriessh/work/fmndb/Masters/Exercise3-Master0.otf -mf /Users/fgriessh/work/fmndb/FontMenuNameDB -shw'
makeotf [Error] Failed to build output font file '/Users/fgriessh/work/fmndb/Masters/Exercise3-Master0.otf'.```

That is not the expected behavior, is it? Can I install the PR differently?

frankrolf:
Same result when using a pointer to the commit instead:
`pip install git+https://github.com/adobe-type-tools/afdko.git@797a6d757381aae2a0be7fc17035b9ae899f1fce`

frankrolf:
Just for being thorough, I also tried creating the virtual environment differently – using `virtualenv`. Same overall result.

I am curious though – the version number fix does not seem to make it through, when doing `pip list`, the afdko lists as `afdko             2.7.1.dev1226+g797a6d75` instead of 2.8.0. Expected?

miguelsousa:
The steps you did should have worked. Not sure why they don't.

punchcutter:
Worked for me, but I ran into something else. I had the name commented out in the `FontMenuNameDB` so this still found it, but couldn't parse it because it was commented out.

miguelsousa:
> I had the name commented out

Good catch. I'll update the PR to account for that case.

frankrolf:
Confirmed working now – just installed the commit without Virtual Env. Thanks!

PR #1175, opened 2020-06-25
[autohint, stemhist, requirements, tests, setup] updates related to psautohint

### LINKED ISSUE #826: Remove autohint tool

### LINKED ISSUE #827: Remove stemhist tool
Remove stemhist tool in favor of psstemhist (parallels autohint -> psautohint transition in #826 )

josh-hadley:
- update requirements to use latest psautohint
- remove autohint.py and related code: closes #826 
- remove stemhist.py and related code: closes #827 
- update AFDKO-Overview.md description of psautohint/psstemhist tools to indicate that they are in a separate repo
- update convertfonttocid.py documentation (autohint -> psautohint)


python/afdko/convertfonttocid.py
```diff
@@ -4,7 +4,7 @@
 convertfonttocid.py. v 1.13.1 Jul 12 2019
 
 Convert a Type 1 font to CID, given multiple hint dict defs in the
-"fontinfo" file. See autohint help, with the "-hfd" option, or the makeotf
+"fontinfo" file. See psautohint help, with the "-hfd" option, or the makeotf
```
miguelsousa, line 7: `-hfd` → `--doc-fddict`
BTW, I didn't get a deprecation message in the console when I used the legacy option.

josh-hadley: Probably [this](https://github.com/adobe-type-tools/psautohint/blob/d1ca70affebe4f070037365d44780ab8a78778b0/python/psautohint/__main__.py#L829) is the cause of the deprecation message not appearing -- it doesn't use `_DeprecatedAction` like the others...possibly because it uses an extended help. I opened https://github.com/adobe-type-tools/psautohint/issues/260 to track. I do not think it is an urgent fix.

docs/AFDKO-Overview.md
```diff
@@ -17,9 +17,11 @@ The tools fall into several different functional groups.
 
 ## 2.1 Making/editing fonts
 
-### `autohint`/`psautohint`
+### `psautohint`
+
+This program is the Adobe auto-hinter and lives in [its own project repo](https://github.com/adobe-type-tools/psautohint). It is used by several ADFDKO tools and is installed automatically via AFDKO's project requirements.
```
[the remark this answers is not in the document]
josh-hadley: Fixed in https://github.com/adobe-type-tools/afdko/pull/1175/commits/11df2a18e8057eebb367f7192861d3b804502006

python/afdko/convertfonttocid.py
josh-hadley: flag descriptor fixed in https://github.com/adobe-type-tools/afdko/pull/1175/commits/567b665afd6d67927896b18f29ceec99c7b6c124

miguelsousa:
Looking at https://codecov.io/gh/adobe-type-tools/afdko/pull/1175/changes it seems that we lost a bit of coverage on the C side. I wonder which removed test made those lines in `c/public/lib/source/cffwrite/cffwrite_fdselect.c` to be hit.

josh-hadley:
> Looking at https://codecov.io/gh/adobe-type-tools/afdko/pull/1175/changes it seems that we lost a bit of coverage on the C side. I wonder which removed test made those lines in `c/public/lib/source/cffwrite/cffwrite_fdselect.c` to be hit.

Yeah, I think it will probably be easier to add new test code for one of the various C tools where the cffwrite is used than to try to track down the specific test.

PR #1192, opened 2020-07-21
[makeotfexe] support OS/2.sFamilyClass in feature files

### LINKED ISSUE #1191: [makeotfexe] support for OS/2 sFamilyClass
[A number of OS/2 fields can be defined in feature files](https://github.com/adobe-type-tools/afdko/blob/develop/docs/OpenTypeFeatureFileSpecification.md#9.f) but it appears it is not possible to define [sFamilyClass](https://docs.microsoft.com/en-us/typography/opentype/spec/ibmfc)?

I am asking because an AppKit developer [filed an issue](https://github.com/IBM/plex/issues/330) on the IBM Plex repo and it seems that [Apple CoreText is providing values from sFamilyClass](https://developer.apple.com/documentation/coretext/ctfontstylisticclass).

josh-hadley:
## Description

Update .fea parsing to accept token `FamilyClass` for OS/2 table definitions (sets the `OS/2.sFamilyClass` field). Syntax is:
`FamilyClass <number>` (inside of a `table OS/2 { ... } OS/2` block). `<number>` can be any decimal, hexadecimal, or octal number from 0 - 65535.

Related: updated a number of OS/2 table field definitions to use `uint16_t` and `int16_t` (intsead of `unsigned short` and `short`) as part of an overall effort to use explicit bit lengths in OpenType tables.

Added/updated tests, documentation, and updated version #s in relevant modules.

Closes #1191


c/makeotf/makeotf_lib/source/hotconv/OS_2.c
```diff
@@ -345,7 +346,7 @@ int OS_2Fill(hotCtx g) {
     h->tbl.yStrikeoutSize = font->win.StrikeOutSize;
     h->tbl.yStrikeoutPosition = font->win.StrikeOutPosition;
 
-    h->tbl.sFamilyClass = CLASS_NONE; /* No classification */
+    // h->tbl.sFamilyClass = CLASS_NONE; /* No classification */
```
cjchapman, line 349: Please remove this commented-out code.

josh-hadley: d'oh! Thanks for catching this. Fixed in https://github.com/adobe-type-tools/afdko/pull/1192/commits/a41e3f954209b80645cc16d1b418be167d92be91

PR #1194, opened 2020-07-22
[docs] update STAT location info [skip ci]

josh-hadley:
## Description

Update spec with corrections to STAT location desriptions:
- use correct value of -32768 for location minimums
- clarify spaces around hyphen for range definition in location format B
- add some internal links

### DIFF
--- a/docs/OpenTypeFeatureFileSpecification.md
+++ b/docs/OpenTypeFeatureFileSpecification.md
@@ -3610,6 +3610,7 @@ flag ElidableAxisValueName OlderSiblingFontAttribute;
 
 The `location` statement takes several formats:
 
+<a name="stat_loc_format_a"></a>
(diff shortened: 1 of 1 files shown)


docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -3610,28 +3610,33 @@ flag ElidableAxisValueName OlderSiblingFontAttribute;
...
 location <axisTag> <nominalValue> <rangeMinValue> - <rangeMaxValue>;
 ```
 
-Format for `axisTag` and the other values as above, and there must be a space
-before the `-` to distinguish it from a possible minus sign for
-`<rangeMaxValue>`. To specify an open ended range use `-32767` to mean 
-negative infinity and `32767.99998` to mean positive infinity. 
+Format for `axisTag` and `value`s is the same as for [location format
+A](#stat_loc_format_a) above. Note that the hyphen (`-`) defining the range must have spaces before
+and after so it does not get treated as a minus sign for `<rangeMaxValue>`.
```
khaledhosny, line 3635: Space before is not required by the current grammar, someone reading this might make an implementation incompatible with makeotf.

josh-hadley:
Closing as we plan to modify the syntax and will update docs accordingly. See https://github.com/adobe-type-tools/afdko/issues/1197

PR #1200, opened 2020-07-30
Add some STAT validation to buildcff2vf

punchcutter:
Check that all fvar axes are defined in the STAT table (required by spec)
Check that STAT axis values are within the ranges defined in the fvar
Update STAT axis name IDs to use the same as fvar (spec says required although no implementations seem to complain yet either way)


tests/buildcff2vf_test.py
```diff
@@ -113,3 +115,83 @@ def test_bug1003_compat():
...
+    """
+    input_dir = get_input_path('STAT_tests')
+    temp_dir = get_temp_dir_path('STAT_tests')
+    copytree(input_dir, temp_dir)
+    ds_path = os.path.join(temp_dir,
+                           'STAT_axis_missing/STAT_axis_missing.designspace')
+    try:
+        runner(CMD + ['-o', 'd', f'_{ds_path}'])
+    except subprocess.CalledProcessError:
+        pytest.xfail()
```
josh-hadley, line 132: Let's use:
```python
with pytest.raises(subprocess.CalledProcessError):
    runner(CMD + ['-o', 'd', f'_{ds_path}'])
```
With try/except, the `runner` could succeed and this test would still pass which is probably not desired.

tests/buildcff2vf_test.py
```diff
@@ -113,3 +115,83 @@ def test_bug1003_compat():
...
+    """
+    input_dir = get_input_path('STAT_tests')
+    temp_dir = get_temp_dir_path('STAT_tests')
+    copytree(input_dir, temp_dir)
+    ds_path = os.path.join(temp_dir,
+                           'STAT_axis_ranges/STAT_axis_ranges.designspace')
+    try:
+        runner(CMD + ['-o', 'd', f'_{ds_path}'])
+    except subprocess.CalledProcessError:
+        pytest.xfail()
```
josh-hadley, line 147: same as above (`pytest.raises`)

tests/buildcff2vf_test.py
```diff
@@ -113,3 +115,83 @@ def test_bug1003_compat():
...
+    """
+    input_dir = get_input_path('STAT_tests')
+    temp_dir = get_temp_dir_path('STAT_tests')
+    copytree(input_dir, temp_dir)
+    ds_path = os.path.join(temp_dir,
+                           'STAT_no_AxisValues/STAT_no_AxisValues.designspace')
+    try:
+        runner(CMD + ['-o', 'd', f'_{ds_path}'])
+    except subprocess.CalledProcessError:
+        pytest.xfail()
```
josh-hadley, line 179: same as above (`pytest.raises`)

python/afdko/buildcff2vf.py
```diff
@@ -484,10 +587,14 @@ def main(args=None):
...
-        varFont.importXML(stat_file_path)
+        import_stat_override(varFont, stat_file_path)
+
+    validate_stat_axes(varFont)
+    validate_stat_values(varFont)
+    update_stat_name_ids(varFont)
 
     varFont.save(options.var_font_path)
-    logger.progress("Built variable font '%s'" % (options.var_font_path))
+    logger.progress("Built variable font '%s'" % options.var_font_path)
```
josh-hadley, line 597: At some point it would be good to update everything to use f-strings. This is fine for now, but the more I use f-strings the more I despise `%`-formatting and `.format()`.

punchcutter: Might as well start updating now.

tests/buildcff2vf_test.py
punchcutter: Yep, little better now.

PR #1203, opened 2020-07-31
Improve test speed

josh-hadley:
## Description

Some minor updates to utilities & test data to improve the speed of test runs:
 - modified logic of `fdkutils.get_font_format()`
 - reduced size of some `otc2otf` test data


miguelsousa:
@josh-hadley It might be worth keeping the `DSIG` and `BASE` tables in the test fonts. It's a small amount of extra data and trigger more occurrences of **-- shared --**

miguelsousa:
Also, I'd expect the `.ttx` files (in the **output** folder) to need to be trimmed as well, since the `.ttc` sources were.

josh-hadley:
> Also, I'd expect the `.ttx` files (in the **output** folder) to need to be trimmed as well, since the `.ttc` sources were.

Ah...that's a good point, that diff should have come up in testing 🤔 

Oh, this is probably why:
https://github.com/adobe-type-tools/afdko/blob/fc8f19770eccfce125cb306c8acf558c7e7361cf/tests/otc2otf_test.py#L102-L107

That should probably be `assert`, not `return`, I think.

josh-hadley:
> @josh-hadley It might be worth keeping the `DSIG` and `BASE` tables in the test fonts. It's a small amount of extra data and trigger more occurrences of **-- shared --**

Agree on the data size, but neither the presence of those two tables nor the occurrences of `--shared--` contribute to coverage (this tool has nearly 100% coverage!), so I'm inclined to leave them out for now. If something changes in that regard, we can add them back then.

miguelsousa:
> That should probably be assert, not return, I think.

Agreed.

miguelsousa:
@josh-hadley How much did things speed up?

josh-hadley:
> @josh-hadley How much did things speed up?

I am gathering some stats on that. It's a bit tricky to measure because of the variability of the various CIs. But I know for the specific case of the otf2otc case, slimming the file sizes cut the time in half.

josh-hadley:
@miguelsousa regarding the speed-up: I gave up on trying to calculate this from the CI stats; there's just too much variability and we don't yet have enough data from the updated suite to make a useful comparison. So I decided to try the test suite on a local machine, repeated 10x (`pytest -q --count=10`) on the current default (after this PR was merged) versus the current master (v3.5.0). Here's the `pytest` stats from those runs:
```
default 2020-08-01: 162 failed, 12568 passed, 40 skipped, 80 xfailed in 1416.36s (0:23:36)
 master 2020-08-01: 162 failed, 12498 passed, 40 skipped, 80 xfailed in 1568.81s (0:26:08)
```
So for the complete test suite, looks like ~10% faster. This isn't super-scientific, and isn't quite an apples-to-apples, because the newer branch has more tests (!), but I think it's close enough to draw a conclusion. If I have some spare time this weekend, I might try run comparisons of the individual tool tests to give a better sense of where the improvements happened.

One other thing I noted is that there are several tests that pass on the first run but fail on subsequent runs -- something we should probably investigate as part of a separate effort. This probably has a small (but ignorable) effect on the times.

miguelsousa:
Thanks for the info.

josh-hadley:
Some additional detail. These are Python apps that use `fdkutils.get_font_format()` (including `otc2otf` which also had its data slimmed), count of 100 this time:
```
pytest -q --count=100 tests/checkoutlinesufo_test.py
default 2020-08-01: 1100 passed in 876.43s (0:14:36)
 master 2020-08-01: 1100 passed in 881.84s (0:14:41)

pytest -q --count=100 tests/makeotf_test.py
default 2020-08-01: 891 failed, 11109 passed in 4284.16s (1:11:24) *
 master 2020-08-01: 891 failed, 11109 passed in 4270.79s (1:11:10)

pytest -q --count=100 tests/otc2otf_test.py
default 2020-08-01: 1900 passed in 222.47s (0:03:42)
 master 2020-08-01: 1900 passed in 1760.43s (0:29:20)

pytest -q --count=100 tests/ttfcomponentizer_test.py
default 2020-08-01: 2300 passed in 117.78s (0:01:57)
 master 2020-08-01: 2200 passed in 118.01s (0:01:58)

pytest -q --count=100 tests/ttfdecomponentizer_test.py
default 2020-08-01: 1500 passed in 38.58s
 master 2020-08-01: 1500 passed in 52.52s

pytest -q --count=100 tests/fdkutils_test.py
default 2020-08-01: 3800 passed, 400 skipped in 17.43s *
 master 2020-08-01: 3800 passed, 400 skipped in 17.27s
```

With two exceptions (and fairly minor diffs there), the new scheme is faster, sometimes significantly so (`otc2otf`, `ttfdecomponentizer`).

PR #1206, opened 2020-08-07
[makeotf] Allow anonymous glyphclass in LookupFlags

khaledhosny:
## Description

Almost everywhere else when a glyph class name is accepted, anonymous glyph classes are also accepted. This is rather inconsistent and inconvenient

## Checklist:


cjchapman:
Looks like Travis passed, but never got back to GitHub about it:
https://travis-ci.org/github/adobe-type-tools/afdko/builds/715929569

khaledhosny:
> Looks like Travis passed, but never got back to GitHub about it:
> https://travis-ci.org/github/adobe-type-tools/afdko/builds/715929569

Something is broken with Travis indeed, I force pushed to re-trigger the build but the same happened again. Build https://travis-ci.org/github/adobe-type-tools/afdko/builds/716378446 passes but it does not report it here.

cjchapman:
@khaledhosny Yes, Travis is definitely broken. We're seeing the same problem in other PRs. Josh is working on using GitHub Actions instead of Travis CI so that we can drop Travis.

khaledhosny:
> 4d-8 and 4d-10 are the exact same. Was one meant to do MarkAttachmentType and the other UseMarkFilteringSet?

Yes, fixed.

PR #1208, opened 2020-08-11
agd: support five-digit code points

### LINKED ISSUE #1207: [comparefamily] [agd] trouble with five-digit code points
I noticed the `comparefamily` report for some code points in Source Serif was suspicious:

```
	Warning. Font uses working name 'u1F16A' rather than Adobe final name 'u1F160'. SourceSerif4-Regular
	Warning. Font uses working name 'u1F16B' rather than Adobe final name 'u1F160'. SourceSerif4-Regular
```
The code points in question are 
- `U+1F16A RAISED MC SIGN 🅪`
- `U+1F16B RAISED MD SIGN 🅫`

comparefamily suggests the final name (and hence the code point) should be `U+1F160 NEGATIVE CIRCLED LATIN CAPITAL LETTER Q 🅠` – wrong, of course.

When investigating, I noticed that the [method for getting the final glyph name](https://github.com/adobe-type-tools/afdko/blob/20b4817b9b1e3c557080f98a66708bb91968fa2a/python/afdko/comparefamily.py#L2752) returns false results for almost all five-digit code points:

```
import os
import unicodedata
from afdko import agd, fdkutils
print(agd.__file__)
resources_dir = fdkutils.get_resources_dir()
kAGD_TXTPath = os.path.join(resources_dir, "AGD.txt")
print('AGD file used:', kAGD_TXTPath)
fp = open(kAGD_TXTPath, "r")
agdTextPath = fp.read()
fp.close()
gAGDDict = agd.dictionary(agdTextPath)

codepoint_start = int('10000', 16)
codepoint_end = int('FFFFF', 16) + 1
for i in range(codepoint_start, codepoint_end):
    hex_value = '{0:5X}'.format(i)
    u_name = 'u' + hex_value

    if gAGDDict.glyph(u_name):
        print(u_name)
        try:
            ud_name = unicodedata.name(chr(i))
            print(ud_name)
        except ValueError:
            print('no unicode name')
        print(gAGDDict.glyph(u_name).name)
        print(gAGDDict.glyph(u_name).uni)  # gets truncated to 4 digits
        print()
```

(this code is mostly pulled from `comparefamily.py`)

Looking through agd.py it seems that five-digit Unicode names are simply not expected:
https://github.com/adobe-type-tools/afdko/blob/20b4817b9b1e3c557080f98a66708bb91968fa2a/python/afdko/agd.py#L47
https://github.com/adobe-type-tools/afdko/blob/20b4817b9b1e3c557080f98a66708bb91968fa2a/python/afdko/agd.py#L63

I fixed this in my local agd.py, and will send a PR in a minute.

frankrolf:
## Description

bug fix to support 5-digit code points in tools relying on the agd.py module
resolves #1207

## Checklist:

- [X] I have followed the [Contribution Guidelines](https://github.com/adobe-type-tools/afdko/blob/develop/CONTRIBUTING.md)
- [X] I have performed a self-review of my own code

I would love to check all the boxes here, but don’t know where to start.
If necessary for this fix I’d appreciate being walked through. Thanks!

### DIFF
--- a/python/afdko/agd.py
+++ b/python/afdko/agd.py
@@ -44,7 +44,7 @@ class glyph:
 		for e in ee:
 			k, x = e # get key and its data
 			if k == 'uni':
-				self.uni = re.compile(r'[0-9A-F]{4}').search(x).group() # Unicode value
+				self.uni = re.compile(r'[0-9A-F]{4,5}').search(x).group() # Unicode value
 			elif k == 'fin': self.fin = re_name.search(x).group() # final name
 			elif k == 'ali':
 				for a in re_name.findall(x): self.ali.append(a) # name aliases
@@ -60,7 +60,7 @@ class glyph:
 	# Check a glyph for internal conflicts, problems, etc.
 	def check(self):
 		m = [] # holds reports of what was fixed
-		u = re.compile(r'uni([0-9A-F]{4})$') # Unicode value regexp
+		u = re.compile(r'u(ni)?([0-9A-F]{4,5})$') # Unicode value regexp
 		n = {}
 		# process aliases: remove duplicates, uninames, final names:
 		for a in self.ali:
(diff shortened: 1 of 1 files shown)


python/afdko/agd.py
```diff
@@ -60,7 +60,7 @@ def parse(self, intext):
 	# Check a glyph for internal conflicts, problems, etc.
 	def check(self):
 		m = [] # holds reports of what was fixed
-		u = re.compile(r'uni([0-9A-F]{4})$') # Unicode value regexp
+		u = re.compile(r'u(ni)?([0-9A-F]{4,5})$') # Unicode value regexp
```
miguelsousa, line 63: You could use `?:` in `(?:ni)?` to make that group not count, which would avoid the group index adjustments further down.

frankrolf: Cool! Didn’t know about that!

PR #1212, opened 2020-08-19
Make designspace attributes lowercase.

### LINKED ISSUE #1211: [makeinstancesufo] wrong camelCasing in designspace attribute
If an instance entry in a designspace file does not contain a PostScript font name attribute, `makeinstancesufo` complains in the following way:
```
WARNING:afdko.makeinstancesufo:Instance at index 0 has no 'postScriptFontName' attribute.
```

However, the ps font name needs to be indicated like this:
```
postscriptfontname="BlahBlah-Blah"
``` 

Note the casing being different in warning vs required.

frankrolf:
Make designspace attributes all lowercase, corresponding to the specification here: https://github.com/fonttools/fonttools/tree/master/Doc/source/designspaceLib\#attributes-9
The spec itself seems to be inconsistent in attribute casing.

Tested with a `<designspace format="3">` file.

Resolves #1211 .

### DIFF
--- a/python/afdko/makeinstancesufo.py
+++ b/python/afdko/makeinstancesufo.py
@@ -241,7 +241,7 @@ def validateDesignspaceDoc(dsDoc, dsoptions, **kwArgs):
     for i, inst in enumerate(dsDoc.instances):
         if dsoptions.indexList and i not in dsoptions.indexList:
             continue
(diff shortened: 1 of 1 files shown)


cjchapman:
Travis CI has been failing to get results back to GitHub, so I'm going to override that particular required status.

PR #1221, opened 2020-09-04
Clarify glyph name ranges (2.g.i)

### LINKED ISSUE #1222: [fea syntax doc] description of glyph name ranges is unclear
The current wording in section 2.g.i regarding ranges is unclear. At several points, a reader can be left scratching their heads wondering what was meant.

PeterCon:
The current description of glyph name ranges is unclear in a few critical points.

Fixes #1222 

## Description

This is a proposed enhancement to the feature file spec documentation. The current
wording in section 2.g.i regarding ranges is unclear. At several points, a reader can be
left scratching their heads wondering what was meant. This is intended to provide
greater clarity in that section.

## Checklist:

- [X ] I have followed the [Contribution Guidelines](https://github.com/adobe-type-tools/afdko/blob/develop/CONTRIBUTING.md)

### DIFF
--- a/docs/OpenTypeFeatureFileSpecification.md
+++ b/docs/OpenTypeFeatureFileSpecification.md
@@ -686,6 +686,18 @@ in an alternate substitution lookup type rule.
 <a name="2.g.i"></a>
 #### 2.g.i. Ranges
 
+A glyph range is a notational mechanism in the feature file grammar that makes it
+possible to define a class of several glyphs is a concise way. The mechanism makes
+use of glyph names that use a contiguous alphabetic sequence A–Z or a–z (or
+sub-sequences thereof), or that use contiguous numeric sequences, such as 0–9. A
+range is specified by referencing starting and ending glyph names, and all of the
+glyph names in the implied sequence are included in the class. The glyphs referenced
+by these names do not have to be in a contiguous sequence in a font file or sources;
+only their names need to be in a contiguous sequence.
+
+If a glyph name within the implied sequence does not correspond to a glyph in the font
+file or font sources, it is ignored.
+
 A range of glyphs is denoted by a hyphen:
 
 ```fea
@@ -703,7 +715,8 @@ feature file glyph names. For example:
 For CID fonts, the ordering is the CID ordering.
 
 For non-CID fonts, the ordering is independent of the ordering of glyphs in the
-font. `<firstGlyph>` and `<lastGlyph>` must be the same length and can differ:
+font. `<firstGlyph>` and `<lastGlyph>` must be the same length and can differ only
+in one of the following ways:
 
 1.  By a single letter from A-Z, either uppercase or lowercase. For example:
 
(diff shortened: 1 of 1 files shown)


frankrolf:
The new examples added imply that glyph ranges can be given without spaces (such as `A-Z`).
This is no longer the case. A glyph range must be specified `A - Z` (with spaces), while `A-Z` is interpreted as a glyph name.

The reason for this change was a font editor’s decision to allow glyph names such as `be-cyr`.

PeterCon:
No examples were added.

frankrolf:
No?
image

PeterCon:
Those are not (intended as) examples of usage or syntax specification but rather are simply describing the characters that are significant for this mechanism. The syntax is stated after this paragraph, and is clear about requiring spaces.

PeterCon:
Revised the change to avoid potential confusion as suggested by @frankrolf .

frankrolf:
Thanks!

khaledhosny:
`A-Z` is interpreted as a glyph name only if the font has a glyph with that name, otherwise it is interpreted as a glyph range.

frankrolf:
I tested my statement with a feature file that has 
```
sub A-Z by A.sc-Z.sc;
```
and received the complaint “font does not contain a glyph named `A-Z`.“
At any rate, I think it’s better to stay consistent?

frankrolf:
I tested my statement with a feature file that has 

```
sub A-Z by A.sc-Z.sc;
```

and received the complaint “font does not contain a glyph named `A-Z`.“
At any rate, I think it’s better to stay consistent?

EDIT: just tried again:

```
makeotfexe [ERROR] <Test> Glyph "A-Z" not in font. [/Users/fgriessh/Desktop/test/features.fea 2]
makeotfexe [ERROR] <Test> Glyph "A.sc-Z.sc" not in font. [/Users/fgriessh/Desktop/test/features.fea 2]
makeotfexe [FATAL] <Test> aborting because of errors
```

khaledhosny:
What I said is described in https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#2fi-glyph-name:
> In order disambiguate whether a text token is a range argument or a glyph name which contains a hyphen, the implementation should first assume that the token is a glyph name. If there is no glyph in the source data with a name that matches the token, then the implementation should interpret the token as a range argument.

If `makeotf` does not actually implement this, then text should be removed from the spec (this would also help simplifying FontTools’ feaLib implementation, which requires a font or glyph list just to disambiguate glyph names as described above).

khaledhosny:
> I tested my statement with a feature file that has
> 
> ```
> sub A-Z by A.sc-Z.sc;
> ```
> 
> and received the complaint “font does not contain a glyph named `A-Z`.“

Ah, yes. There is no ambiguity here as a range can’t possibly work. There range needs to be in a glyph class:
```fea
sub [A-Z] by [A.sc-Z.sc];
```

frankrolf:
Thanks for the clarification! I agree allowing both notation formats creates ambiguity.

PeterConstable:
Just noting: the issue you're discussing is orthogonal to the one my PR addresses.

frankrolf:
That’s correct, however I think the change you made is good (thanks again!).
I will leave the acceptance of the PR up to higher powers, since I don’t want to overrule whatever concerns others may have.
(Since it’s labor day in the US today, we probably won’t get another reaction until tomorrow).
Thanks for your patience.

josh-hadley (approved):
:+1: Thank you!

PR #1228, opened 2020-09-24
Revert #1212, lowercase error message only

### LINKED ISSUE #1211: [makeinstancesufo] wrong camelCasing in designspace attribute

frankrolf:
This really fixes #1211.
The attributes stay camelCase, only the error message is reported in all-lowercase.

## Checklist:

- [X] I have followed the [Contribution Guidelines](https://github.com/adobe-type-tools/afdko/blob/develop/CONTRIBUTING.md)
- [X] I have verified that new and existing tests pass locally with my changes
- [X] I have performed a self-review of my own code

### DIFF
--- a/python/afdko/makeinstancesufo.py
+++ b/python/afdko/makeinstancesufo.py
@@ -241,10 +241,11 @@ def validateDesignspaceDoc(dsDoc, dsoptions, **kwArgs):
     for i, inst in enumerate(dsDoc.instances):
         if dsoptions.indexList and i not in dsoptions.indexList:
             continue
-        for attr_name in ('familyname', 'postscriptfontname', 'stylename'):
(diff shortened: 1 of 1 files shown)


josh-hadley (approved):
Looks good, thanks @frankrolf!

frankrolf:
Thanks, and sorry for messing up! :-)

PR #1231, opened 2020-10-02
[checkoutlinesufo] fix nested loop variable

cjchapman:
This is a fix for the following issue identified by LGTM:
```
Nested for statement uses loop variable 'point' of enclosing for statement.
```

(I created this PR mainly to check if our LGTM configuration is still working now that GitHub's officially released "code scanning".)


cjchapman:
So this ended up being more informative than I expected, in that it:
* confirmed that LGTM is still running on our PRs
* revealed a bug in tx_test.py that is fixed in https://github.com/adobe-type-tools/afdko/pull/1232

josh-hadley (approved):

Issue #1245, opened 2020-10-26
[tx] (t1w) glyph type doesn't match font type


takaakifuji:
```
$ tx -t1 -o font.pfa font.ufo
```
```
tx: --- font.ufo
tx: (t1w) glyph type doesn't match font type
```
`tx` started to fail when `font.ufo` contains glyph names in the `cid00000` format. It seems that the change in fb27836 causes this failure.

punchcutter:
I guess the UFO contain a mix of glyphs with `cid00000` format names and non CID names? The assumption was that all glyphs in a CID-keyed font would be named with the `cid00000` pattern and otherwise wouldn't be named like that, but I see now it's not the best assumption. We will adjust it to account for this.

takaakifuji:
As far as I observe the issue with the `3.5.1` release, `tx` fails when at least one glyph is named with `cid00000` format in the given UFO.

Talking of `tx`, I suppose the output of `$ tx -t1` should always be name-keyed anyway. My concern is that such behavioral change of `tx` could potentially break the existing build pipelines for CJK fonts.

punchcutter:
This fix was meant only for UFO because `tx` has never correctly handled UFO as a CID-keyed source. The change shouldn't break existing build pipelines because that part never worked in the first place. It should only be a problem right now because of the mistaken logic in reading `cidxxxxx` names.

takaakifuji:
After I posted to #1247, I managed to have a quick look at the codebase.

`t1write.c` ensures that the CIDFontName and the ROS are not empty:
https://github.com/adobe-type-tools/afdko/blob/6b810de99b31e6e9c85331bb6a006b5603cd246e/c/public/lib/source/t1write/t1write.c#L1899-L1903

My take is to evaluate the same condition in prior to testing the name of each glyph in `uforead.c`:
```c
if (h->top.cid.Registry.ptr != ABF_UNSET_PTR &&
    h->top.cid.Ordering.ptr != ABF_UNSET_PTR &&
    h->top.cid.Supplement != ABF_UNSET_INT &&
    h->top.cid.CIDFontName.ptr != ABF_UNSET_PTR) { ... }
```
https://github.com/adobe-type-tools/afdko/blob/6b810de99b31e6e9c85331bb6a006b5603cd246e/c/public/lib/source/uforead/uforead.c#L3455-L3466

`tx` complains if:

1. `ABF_CID_FONT` has empty CIDFontName or ROS
[...]

PR #1247, opened 2020-10-26
Fix reading cid names in UFOs

punchcutter:
Fix for https://github.com/adobe-type-tools/afdko/issues/1245
We only set CID flags if all glyphs conform to the same `cidXXXXX` naming pattern. Otherwise we can't assume it is meant to be a CID-keyed font.

### DIFF
--- a/c/public/lib/source/uforead/uforead.c
+++ b/c/public/lib/source/uforead/uforead.c
@@ -3712,7 +3712,7 @@ int ufoIterateGlyphs(ufoCtx h, abfGlyphCallbacks* glyph_cb) {
     for (i = 0; i < h->chars.index.cnt; i++) {
         hasCIDNames(h, i, glyph_cb);
     }
-    
+
     for (i = 0; i < h->chars.index.cnt; i++) {
         int res;
         res = readGlyph(h, i, glyph_cb);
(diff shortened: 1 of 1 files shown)


josh-hadley (approved):
👍

miguelsousa:
Does this need a test case?

punchcutter:
@miguelsousa Yes, I was planning to add some test cases, but wanted to get the fix up quickly in case @takaakifuji would like to use it now.

takaakifuji:
Thank you for compiling the pull request!

Well, this patch expects that `$ tx -t1 -o font.pfa font.ufo` should fail if every single glyph in the input font is named with `cid00000` convention, is that correct? Comparing to the 3.5.0 release, that sounds a bit like a regression to me, because one of our usage of `tx -t1` (often along with `-g` option) is to break up a whole UFO into multiple Type 1 subsets so that they should be compiled as FDArray elements afterwards. Is there any workaroud to reliably force convert a UFO to Type 1 regardless of the glyph naming scheme? I'm afraid some change are also needed for `tx.c` so that it can always forcibly treat the input font as name-keyed when `-t1` option is given, even in case it is flagged as CID-keyed...I'm not in hurry at all, and I think it's better to spend more time to settle this down.

As for this patch, I think there's also an option to check the existence of the `com.adobe.type.cid.*` keys in `fontinfo.plist`. I personally couldn't catch up with the background scenario for the whole change yet, but that might be more natural and the intention might become clearer rather than guessing by the glyph names.

punchcutter:
This needs a lot more work which is now in progress on a separate branch so closing this PR.

Issue #1234, opened 2020-10-05
Things fea can't do: delete a glyph


simoncozens:
In my workshop I've been telling people "You can't delete a glyph using a substitution rule". The OT spec says as much: "A Multiple Substitution (MultipleSubst) subtable replaces a single glyph with more than one glyph". *But*... [you can](https://bugzilla.mozilla.org/show_bug.cgi?id=1128658), Uniscribe, Harfbuzz and CoreText all support it, and Arial Unicode relies on the fact in its Devanagari implementation.

So it would be nice if there were syntax for it.

khaledhosny:
A possible syntax might be `sub foo by <NULL>;`.

simoncozens:
In the workshop I suggested `sub a by;`

khaledhosny:
> In the workshop I suggested `sub a by;`

That was the first to come on my mind, but it feels like a typo. `<NULL>` is used in several places already to mean _nothing to see here, move on_.

anthrotype:
yes, that looks like a typo to me, I prefer Khaled's proposal to use `<NULL>`.

simoncozens:
Ok, I’ll work on a patch.

josh-hadley:
Has anyone proposed a [spec change](https://github.com/MicrosoftDocs/typography-issues) for this yet? It would be nice to have the spec reflect reality and avoid having to explain why AFDKO supports something that is expressly prohibited (even if it is widely supported in implementations).

mbutterick:
This proposal will break every tool that parses `fea` files.

alerque:
@mbutterick No. It will mean old tooling that doesn't get updated won't be interoperable with tooling that does *on projects that use this as a feature*. It won't automatically break old tools across the board or even reduce interoperability with new tools that don't happen to use the new feature.

simoncozens:
It’s even less true than that. Something which is currently a syntax error will become valid. Nothing which is currently valid will become invalid.

mbutterick:
This is my last comment. Your point is true in the sense of concrete instances. But it is not true in the sense of abstract implementations, which is how tools interoperate. As a software author, I read a specification not just to learn what *can* happen but also what can *never* happen (and write tests accordingly). This is why, say, Unified Font Object is [versioned](https://unifiedfontobject.org/): so that tool authors can make stable claims about their code.

simoncozens:
The AFDKO spec is also versioned.

moyogo:
[...]

PR #1251, opened 2020-11-03
Allow deleting glyphs

simoncozens:
This addresses #1234﻿


josh-hadley (changes requested):
@simoncozens thanks for this. This looks good to me, but I'd like to have you rebase & make any adjustments needed to get tests passing before we get it merged, if you don't mind.

simoncozens:
Those aren't my failing tests...

josh-hadley (approved):
Thanks again @simoncozens , looks good and will get this merged.
Looks like we need to iron out some issues with GitHub Actions running tests on forks. We've made a lot of updates to that whole thing since this PR was started/in Draft mode 3 months ago but there are still some details to work out.

PR #1272, opened 2020-12-16
[requirements] remove mutatorMath, update fontTools

miguelsousa:
[`mutatorMath`](https://github.com/LettError/MutatorMath) is a dependency of [`ufoProcessor`](https://github.com/LettError/ufoProcessor) but it's not a direct dependency of **afdko**.


alerque (dismissed):
This is correct. I think it may have once been an optional feature dependency in ufoProcessor which would have made it necessary to specify here, but that is not now the case (if it ever was).

josh-hadley (approved):
LGTM

PR #1279, opened 2021-01-14
[makeotfexe] Allow negative internal leading

### LINKED ISSUE #1227: Any reason why winAscent + winDescent can’t be smaller than the UPM?
https://github.com/adobe-type-tools/afdko/blob/5778828e0c1ac3c425a9c92c77f4a5bacf0d32bc/c/makeotf/makeotf_lib/source/hotconv/hot.c#L716-L721
This might be a good check for auto calculated winAscent + winDescent values. But quite confusing for manually set values.

kaydeearts:
Also updates expected output for makeotf_test.py and makeotfexe_test.py

Fixes #1227


josh-hadley (approved):
Thanks @kaydeearts !

PR #1284, opened 2021-01-26
re-format spot docs

frankrolf:
I re-formatted the docstring in `spot`. I know that it needs to be compiled to make it “real”, but I hope this can just happen with the next FDK-build.

I hope the line length is OK like that.


kaydeearts:
I need to change my test for it to ignore version numbers, so I'll let you know when I'm done for you to rebase.

kaydeearts:
I've fixed my test case in PR #1286 and it's merged so you can rebase off of that.

frankrolf:
see #1287

PR #1285, opened 2021-01-27
Sync master to develop

josh-hadley:
Syncing up a backlog of commits to `master` that did not ever get added to `develop`, leaving them out of sync. This will also ensure that release tags are present on `develop`.


cjchapman (commented):
This looks good. I confirmed local experiments that this will get the release tags on the `develop` branch.

This is the command I used to check the tags on the current branch:
```bash
git log --decorate --pretty=oneline | grep "(tag:"
```

PR #1289, opened 2021-02-01
Add ignore-contour-order flag to checkoutlinesufo

punchcutter:
In some situations we have no use for attempting to restore the contour order so force it to be skipped with the `--ignore-contour-order` flag.


kaydeearts (approved):
👍  Looks good to me!

Issue #1161, opened 2020-06-10
[makeinstancesufo] performance


frankrolf:
After some internal discussion, and my dreaming in https://github.com/adobe-type-tools/afdko/issues/1160#issue-636400380 , here is an official issue:

Improving the speed of `makeinstancesufo` would be very welcome. When kicking off a typical instance generating session, it’s usually time for a shower and/or coffee. I can only drink a limited amount of coffee! ☕️

PR #1293, opened 2021-02-02
[makeinstancesufo.py] Adding multiprocessing to run

kaydeearts:
This is part of addressing #1161 by adding multiprocessing


josh-hadley (approved):
👍  Looks good and should be a nice improvement for running `makeinstancesufo`!

PR #1294, opened 2021-02-03
Add "sub A by NULL" to specification

simoncozens:
See #1234 

Note that this is *described* as a 5.a single substitution - because the feature syntax is "`sub <something> by <something else>`" - even though (as an implementation detail) it compiles to a multiple substitution.


josh-hadley (approved):
Thanks @simoncozens !

PR #1296, opened 2021-02-04
[checkoutlinesufo] Fix restore_contour_order fail

### LINKED ISSUE #1291: [checkoutlinesufo] restore_contour_order fails in some situations
In some outlines like the glyph in tests/checkoutlinesufo_data/input/contour-restore.ufo `checkoutlinesufo` fails to restore contour order because of the way the contours overlap by a single point. 

Run
`checkoutlinesufo -e --all contour-restore.ufo`

and the result is

`AssertionError: Failed assertion: duplicated start point on contour 3 at 616, 597 of glyph cid51107.`

This stops the entire process, but we should never fail here. The main purpose of `restore_contour_order` is to attempt to keep compatibility the same across masters, but that's not always going to be possible, especially when removing overlaps.  If we can't restore contour order then we should just report it.

kaydeearts:
Fixes #1291

### DIFF
--- a/python/afdko/checkoutlinesufo.py
+++ b/python/afdko/checkoutlinesufo.py
@@ -943,20 +943,19 @@ def restore_contour_order(fixed_glyph, original_contours):
                                 if (point.x == old_start_point.x) \
                                         and (point.y == old_start_point.y) \
                                         and point.segmentType is not None:
-                                    try:
-                                        msg = ("Failed assertion: duplicated "
-                                               f"start point on contour {ci} "
-                                               f"at {point.x}, {point.y} of "
-                                               f"glyph {fixed_glyph.name}.")
-                                        assert not ctr_starts[ci], msg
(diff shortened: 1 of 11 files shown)


josh-hadley (changes requested):
Can you add a test case that checks for the Warning message when `checkoutlinesufo` is run on `contour-restore.ufo`? The current test case that uses that file doesn't hit this code (no warning is emitted) because it uses the `ignore-contour-order` flag.

 You can probably just copy that case (`test_ignore_contour_order`) but remove the `=ignore-contour-order` from the options to `runner()`. Basically just reproduce the command @punchcutter mentions in #1291 and verify that the new Warning message is emitted ("`Warning: duplicated start point on contour 3 at 616, 597 of glyph cid51107.`").

josh-hadley (approved):
👍  Looks good, thanks!

PR #1311, opened 2021-02-22
[makeotfexe] increase uvName string length limit

### LINKED ISSUE #1310: Extend string length limitation for 3rd column of GOADB
Currently, `makeotfexe` has a limitation of 63 characters on the 3rd column of the GOADB (used to map to Unicode values). In [some cases](https://github.com/adobe-type-tools/afdko/issues/1308#issuecomment-781697902) this can result in `makeotfexe` rejecting multiple mappings because the length of the string used to describe multiple mappings exceeds 63 characters.

The length limit should be expanded to allow longer strings and more UV mappings.

josh-hadley:
## Description

Increase string length limit for parsing the 3rd column of a GOADB file (aliases/Unicode Value mappings).

Closes #1310 

## Checklist:


tests/makeotfexe_test.py
```diff
@@ -706,3 +706,31 @@ def test_negative_internal_leading_bug1227():
...
+                 'f', f'_{get_input_path(input_filename)}',
+                 'o', f'_{actual_path}',
+                 'gf', f'_{get_input_path(goadb_filename)}']
+
+    stderr_path = runner(cmd)
+
+    with open(stderr_path, 'rb') as f:
+        output = f.read()
+
+    assert ((b'(record skipped)(gnameError)') in output) is expected
```
cjchapman, line 736: Minor suggestion -- there's a superfluous set of parentheses in this line, and it could be simplified to:
```python
    assert (b'(record skipped)(gnameError)' in output) is expected
```

josh-hadley: ((indeed!))
fixed in https://github.com/adobe-type-tools/afdko/pull/1311/commits/574ed0521376bcb8b00e41822a680770b4fd7a29

kaydeearts (approved):
👍

PR #1314, opened 2021-02-24
[buildcff2vf] clarify reason for VarLibCFFPointTypeMergeErrors

### LINKED ISSUE #1312: buildcff2vf should clarify reason when raising VarLibCFFPointTypeMergeError
When fontTools fails to merge and raises `VarLibCFFPointTypeMergeError` it can happen in `fontTools.varLib.cff` `add_point`, `add_hint`, or `add_hintmask`, but `buildcff2vf` reports the same error:

`The input set requires compatibilization. Please try again with the -c (--check-compat) option.`

Even when `buildcff2vf` is already being run with `-c` we get this same error which is not quite correct. Running the underlying code shows the real error which comes from the hinting. For example

`fontTools.varLib.errors.VarLibCFFPointTypeMergeError: Glyph '.notdef': 'rmoveto' at point index 0 in master index 1 differs from the default font point type 'hstemhm'`
or
`fontTools.varLib.errors.VarLibCFFPointTypeMergeError: Glyph 'cid00005': 'hintmask' at point index 5 in master index 1 differs from the default font point type 'rrcurveto'`

josh-hadley:
## Description

Improve reporting of underlying problem when `VarLibCFFPointTypeMergeError` is encountered in `buildcff2vf`

Closes #1312 

## Checklist:

### DIFF
--- a/python/afdko/buildcff2vf.py
+++ b/python/afdko/buildcff2vf.py
@@ -35,6 +35,7 @@ class CFF2VFError(Exception):
 class CFF2VFCompatibilizationError(VarLibCFFPointTypeMergeError):
     """To distinguish when raised during compatibilization"""
 
+
 # set up for printing progress notes
 def progress(self, message, *args, **kws):
     # Note: message must contain the format specifiers for any strings in args.
--- a/python/afdko/buildcff2vf.py
+++ b/python/afdko/buildcff2vf.py
@@ -32,6 +32,9 @@ class CFF2VFError(Exception):
     """Base exception for buildcff2vf"""
 
 
+class CFF2VFCompatibilizationError(VarLibCFFPointTypeMergeError):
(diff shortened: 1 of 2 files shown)


punchcutter (commented):
I don't see where CFF2VFCompatibilizationError is being used and this still says "try again with the -c option" even when the -c option is used. 

Maybe we can check for  
`if not options.check_compatibility:`
and only say "Please try again" in that case, but even then if it's failing because of hints we know that from the message. Or maybe fontTools can change the hint exceptions. I opened an issue about it: https://github.com/fonttools/fonttools/issues/2197

josh-hadley:
Once https://github.com/fonttools/fonttools/pull/2216 gets merged & packaged in fontTools, I'll adjust and re-open this PR.

josh-hadley:
No action on this, closing.

PR #1318, opened 2021-02-26
[checkoutlinesufo] Fix incorrect logic

### LINKED ISSUE #1315: [checkoutlinesufo] duplicated start point check reports useless information
In `restore_contour_order` the check for duplicate start points warns for tons of glyphs in my test font, but the information isn't useful. It warns of duplicate start points in glyphs that do not contain duplicate start points. For example in this simple glyph the first contour start point remains as the start point when overlaps are removed

uni30A7

uni30A7-overlapremoved

In this case (and many like it) there are no contour orders to restore. The result is a single contour. If the result is a single contour I would expect the first contour's start point to remain as the the start point like it does, but without any warnings. 

Even in glyphs with more contours the warning isn't useful. Here the start points on contour 18, 15, 5, and 10 are all reported as duplicate start points. Before 02cb7fd356807ece7b742d95bad21369581c4ea7 the same glyphs do not have any "duplicate start point" warnings in `restore_contour_order`.  

cid22144

cid22144-overlapremoved

Command `checkoutlinesufo --all test-checkoutlinesufo.ufo -e`
[test-checkoutlinesufo.ufo.zip](https://github.com/adobe-type-tools/afdko/files/6047859/test-checkoutlinesufo.ufo.zip)

kaydeearts:
This should fix #1315 
@punchcutter reported that this error didn't appear before 02cb7fd. 
Within that merge, we had replaced:
`assert not ctr_starts[ci], msg` with
`elif not ctr_starts[ci]:`, when it should've been `elif ctr_starts[ci]:`.


josh-hadley (changes requested):
Looks good but I think we should add a test case (@punchcutter 's ufo attached to #1315 should work, I think) to help avoid regressions on this.

josh-hadley (approved):
Looks good, thanks!

PR #1331, opened 2021-03-12
Moved list sort out of loop to not waste time sorting.

punchcutter:
In a 65k font this made `checkoutlinesufo` spend 15% of the time sorting the list on each iteration.


kaydeearts (approved):
👍 Looks good!

PR #1336, opened 2021-03-22
Spec 6.4.iii: 'markClass' instead of 'mark' in example 4

moyogo:
## Description

The example 4 of section 6.h.iii starts with `mark sukun <anchor 0 0> @TOP_CLASS;` but that should be `markClass sukun <anchor 0 0> @TOP_CLASS;`

## Checklist:

N/A


punchcutter (approved):
Nice catch. 👍

PR #1340, opened 2021-04-11
Fixes and improvements to GH Actions

miguelsousa:
Makes a few fixes and improvements to the Test and Publish workflows, now that the whole infrastructure is a bit more mature.

References that may help with reviewing these changes:
- https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
- https://github.com/pypa/gh-action-pypi-publish
- https://github.com/softprops/action-gh-release
- https://github.com/actions/create-release

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -8,7 +8,6 @@ on:
 jobs:
   build_wheels:
     name: Build Py3 Wheel on ${{ matrix.os }}
-    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


miguelsousa:
This PR needs more work. The release should be created only after all the wheels have been built.

At the end of each platform run the build artifacts should be uploaded. Afterwards the job that creates the release will download the whole set of wheels and publish them to PyPI and GH. Documentation on storing workflow data as artifacts https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts

miguelsousa:
I think this PR is now ready for review. I've tested the new changes by using this repo https://github.com/miguelsousa/test-workflows/

These are some more references I had to use:
- https://github.com/actions/upload-artifact
- https://github.com/actions/download-artifact
- https://github.com/actions/checkout
- https://zxq9.com/archives/795
- https://unix.stackexchange.com/questions/48101/how-can-i-have-date-output-the-time-from-a-different-timezone
- https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

josh-hadley (approved):
Great, thanks for getting this updated -- should be useful for other projects as well.

N.B. I saw that the Coverage-related tests are failing, I believe that's because they're picking up the newer fontTools release upon install and failing for the same reasons as https://github.com/adobe-type-tools/afdko/pull/1341 (TTX-related). I think @kaydeearts has some updates in the works for those so we should be able to get this all sorted out soon.

kaydeearts (approved):
👍  Rebased & now tests are passing. Josh had given an approval before the rebase as well, so now it's good to go.

PR #1344, opened 2021-04-15
[cffread.c] data type fix + add, update tests

### LINKED ISSUE #1343: [tx] t1read producing bad values for FontBBox and BlueValues
Introduced in https://github.com/adobe-type-tools/afdko/commit/2e19f286fdc7497d1edc9a2ac68e56cefbd6e05f

kaydeearts:
Should fix #1343


punchcutter (approved):
Tested with the files we ran into this with and it looks good. Test failing is a separate issue we know about so I think merging this is fine.

PR #1347, opened 2021-04-19
Add error for fonts surpassing 65000 SID Limit

### LINKED ISSUE #1338: [rotateFont] Segmentation Fault happens when using rotateFont or mergeFont to rename glyph names.
### Environment
macOS 11.2.3
python 3.9.2 
AFDKO installed by `pip3 install afdko`

### Command
```
tx -t1 -n -decid -usefd 0 SourceHanSans-Regular.otf decidfont.ps
rotateFont -t1 -rtf map.txt decidfont.ps gidfont.ps
```
[SourceHanSans-Regular.otf](https://github.com/adobe-fonts/source-han-sans/blob/release/OTF/Japanese/SourceHanSans-Regular.otf)
The contents of `map.txt`
```
.notdef	.notdef	None	0	0
cid1	uni0020	None	0	0
```

kaydeearts:
This should fix #1338 
In the bug above, the user ran into segfault error after using rotateFont. This was actually due to the source font having too many glyphs for it to be a name-keyed font (should be CID-keyed font instead).
We added an error for this to inform the user, stop the program & avoid segfault.


c/public/lib/source/t1read/t1read.c
```diff
@@ -55,6 +55,7 @@ static const char *keys[] =
 
 typedef unsigned short STI; /* String index */
 #define STI_UNDEF 0xffff    /* Undefined string index */
+#define STI_LIMIT 65000
```
josh-hadley, line 58: Is the limit 65000 or 65534?

josh-hadley (changes requested):
Question on the actual limit value, see comments. Also: seems like we should probably document this limit and the behavior somewhere (either in tools built-in help or elsewhere in AFDKO documentation)

c/public/lib/source/t1read/t1read.c
cjchapman: It's 65000 -- see section "10 String INDEX" of:
https://wwwimages2.adobe.com/content/dam/acom/en/devnet/font/pdfs/5176.CFF.pdf

Specifically:
> An SID is defined as a 2-byte unsigned number but only takes values in the range 0-64999, inclusive.  SID values 65000 and above are available for implementation use.

josh-hadley (approved):
👍

PR #1350, opened 2021-04-22
[spec] Variable FEA Syntax

josh-hadley:
## Description

Includes the following changes to the OpenType Feature File Specification aka "FEA Syntax":
- proposed syntax for variable scalar values from @simoncozens
- proposed syntax for featureVariations from @punchcutter with additional modifications based on offline discussions

## Checklist:

### DIFF
--- a/docs/OpenTypeFeatureFileSpecification.md
+++ b/docs/OpenTypeFeatureFileSpecification.md
@@ -6,12 +6,12 @@ layout: default
 OpenType™ Feature File Specification
 ---
 
-Copyright 2015-2020 Adobe. All Rights Reserved. This software is licensed as
+Copyright 2015-2021 Adobe. All Rights Reserved. This software is licensed as
 OpenSource, under the Apache License, Version 2.0. This license is available at:
 http://opensource.org/licenses/Apache-2.0.
 
-Document version 1.25.1
-Last updated 5 July 2020
+Document version 1.26.0
+Last updated XX XXXXXX 2021  **TODO: adjust date**
 
 **Caution: Portions of the syntax unimplemented by Adobe are subject to change.**
 
@@ -48,6 +48,7 @@ Last updated 5 July 2020
   - [c. parameters](#4.c)
   - [d. lookupflag](#4.d)
   - [e. lookup](#4.e)
+    - [i. Specifying FeatureVariations (variable fonts)](#4.e.i)
   - [f. markClass](#4.f)
   - [g. subtable](#4.g)
   - [h. Examples](#4.h)
@@ -261,6 +262,9 @@ The following are keywords only in their corresponding table/feature blocks:
 | [`location`](#9.i) | STAT table | ✅ |
 | [`ElidableAxisValueName`](#9.i) | STAT table | ✅ |
 | [`OlderSiblingFontAttribute`](#9.i) | STAT table | ✅ |
+| [`conditionset`](#X.x) TODO: link up | Variable sub and pos lookups | ❌ |
+| [`variation`](#4.e.i) | Variable sub and pos lookups | ❌ |
+
 
 The following are keywords only where a tag is expected:
 
@@ -284,7 +288,7 @@ dflt  # can be used only with the language keyword and as the language value wit
     { }  braces           Enclose a feature, lookup, table, or anonymous block
     [ ]  square brackets  Enclose components of a glyph class
     < >  angle brackets   Enclose a device, value record, contour point, anchor, or caret
-    ( )  parentheses      Enclose the file name to be included
+    ( )  parentheses      Enclose the file name to be included. Within a `<metric>`, enclose variable scalar values.
 
 
 <a name="2.e"></a>
@@ -309,6 +313,27 @@ value records [§[2.e.iv](#2.e.iv)] for positioning rules, as well as to express
 the values of various table fields [§[9](#9)].
 
 _[ Note: Multiple master support has been withdrawn as of OpenType specification 1.3. ]_
+
+_[ Note: the following is unimplemented and is subject to change. ]_
+
+For **variable fonts only**, a `<metric>` value can make use of the following syntax to specify how values should vary based on axis locations:
+```
+(<location_spec1>:<value1> <location_spec_2>:<value2>)
+```
+
+Notes:
+ - A `<location_spec>` consists of one or more `<axis_tag>=<axis_location>` pairs, separated by commas 
+ - `<location_spec>:<value>` pairs are separated by spaces within the variable `<metric>`
+
+Example showing a simple single-axis variable scalar value. The value is -100 when Weight is 200, and -150 when Weight is 900:
+```
+<(wght=200:-100 wght=900:-150)>
+```
+
+A more complex example combining single and multiple `<axis_location>`s extends the above example to specify that the value should be -120 when Weight is 900 and Width is 150:
+```
+<(wght=200:-100 wght=900:-150 wght=900,wdth=150:-120)>
+```
 
 <a name="2.e.iii"></a>
 #### 2.e.iii. Device table _[ Currently not implemented. ]_
@@ -1266,6 +1291,44 @@ the same order that they are written in the feature file. For non-contextual
 rules, the implementation sorts the rules to avoid conflict; for example, the
 ligature substitution rule for f_f_i will be written before the ligature
 substitution rule for f_i, no matter what their order is in the feature file.
+
+<a name="4.e.i"></a>
+#### 4.e.i Feature Variations (variable font feature replacement)
+
+For variable fonts *only*, it is possible to specify [FeatureVariations](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#featurevariations-table), which allow the use of different lookups for different parts of the variation space.
+
+<a name="4.e.i.1"></a>
+##### 4.e.i.1 `conditionset`
+
+A named `conditionset` defines [conditions](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#conditionset-table) that can trigger a variable font lookup to swap in different lookups when the conditions are matched. A `conditionset` is simply a list of axis tags and min/max values. A conditionset is matched when all conditions are true. Syntax:
+```fea
+conditionset <name of set> {
+    <axis tag> <minValue> <maxValue>;
+    # additional tag min max entries
+} <name of set>;
+```
+For example, to define a "heavy" condition, for the 'wght' axis between 700-900:
+```fea
+conditionset heavy {
+    wght 700 900;
+} heavy;
+```
+
+<a name="4.e.i.2"></a>
+##### 4.e.i.2 `variation` (FeatureVariation)
+A `variation` is akin to a regular `feature` definition, with an added `conditionset` specifier to indicate that the feature should be active when the conditions of the conditionset are met. The syntax is:
+```fea
+variation <feature tag> <conditionset name or NULL> {
+    # feature specifications
+} <feature tag>;
+```
+For example, let's say we've defined a `lookup` called "heavy_symbols" that we want to swap in when the "wght" axis is between 700-900 (our "heavy" `conditionset` from above):
+```
+variation rvrn heavy {
+    lookup symbols_heavy;
+} rvrn;
+```
+
 
 <a name="4.f"></a>
 ### 4.f. markClass


simoncozens:
Thanks for this. My part looks good, with the note I've added.

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -1267,6 +1293,44 @@ rules, the implementation sorts the rules to avoid conflict; for example, the
...
+conditionset heavy {
+    wght 700 900;
+} heavy;
+```
+
+<a name="4.e.i.2"></a>
+##### 4.e.i.2 `variation` (FeatureVariation)
+A `variation` is akin to a regular `feature` definition, with an added `conditionset` specifier to indicate that the feature should be active when the conditions of the conditionset are met. The syntax is:
+```fea
+variation <feature tag> <conditionset name or NULL> {
```
anthrotype, line 1322: what's the meaning of `NULL` here?

anthrotype: maybe it means this (quoting from [ot spec](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#featurevariations-table)):

> If a given condition set contains no conditions, then it matches all contexts, and the associated feature table substitution is always applied

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -1267,6 +1293,44 @@ rules, the implementation sorts the rules to avoid conflict; for example, the
...
 
+<a name="4.e.i"></a>
+#### 4.e.i Feature Variations (variable font feature replacement)
+
+For variable fonts *only*, it is possible to specify [FeatureVariations](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#featurevariations-table), which allow the use of different lookups for different parts of the variation space.
+
+<a name="4.e.i.1"></a>
+##### 4.e.i.1 `conditionset`
+
+A named `conditionset` defines [conditions](https://docs.microsoft.com/en-us/typography/opentype/spec/chapter2#conditionset-table) that can trigger a variable font lookup to swap in different lookups when the conditions are matched. A `conditionset` is simply a list of axis tags and min/max values. A conditionset is matched when all conditions are true. Syntax:
```
anthrotype, line 1304: > A conditionset is matched when all conditions are true.

So the `conditionset` statement can only do boolean `AND` but not `OR`. I wonder if this (or another keyword) should be allowed to define not only "boxes" but also "regions", to borrow @justvanrossum's terminology from 
https://github.com/fonttools/fonttools/blob/9959916c64c5ab24de416aabf05c2bdace610014/Lib/fontTools/varLib/featureVars.py#L22-L29

justvanrossum: Haven't followed this, but if .fea strictly follows the OTF FeatureVariations structure, it will end up fairly limited in it's practical use. 

One of the reasons fontTools.varLib.featureVars became so complex* is that it allows both AND and OR.

*) so complex that Behdad rewrote it almost from scratch, to make it behave and perform decently, after my initial sloppy implementation

simoncozens (approved):

docs/OpenTypeFeatureFileSpecification.md
twardoch: What are the situations when "OR" cannot be replaced by separate conditionsets?

twardoch: E.g. `(a OR b) AND (c AND d)` can be rewritten as 

- `a AND c AND d`
- `b AND c AND d`

I may be missing something but if you want 'OR' matching, you just define multiple conditionsets, each describing all 'AND' conditions, no?

twardoch: In FontLab 7, when you assign a special tag to the target glyph, it replaces the source glyph under specified conditions:

```
[srcglyph]~cond1[&cond2][&cond3...]
```

if `srcglyph` is absent, the current glyph’s name without a suffix is used.

Each cond is a 2-letter (internal FL) axis tag with bounds, like `wt>500` or `340>wd` or `100<op<600` etc. — and they're concatenated with `&`.

If I need 'OR', I add a second special tag. 

I know there is some redundancy in this, but I’m genuinely curious if there are conditionsets that involve both OR and AND which cannot be expressed as a list of 'AND' conditionsets.

docs/OpenTypeFeatureFileSpecification.md
josh-hadley: Right, `NULL` here is for "a condition set with no conditions". Although I suppose we could push that back and require a `conditionset` here, even if it is:
```
conditionset empty {
} empty;
```
Do you have a preference?

moyogo: What's the difference between the two following?
```
variation <feature tag> NULL {
# feature specifications
} <feature tag>;
```
and
```
feature <feature tag> {
# feature specifications
} <feature tag>;
```

docs/OpenTypeFeatureFileSpecification.md
simoncozens: I think Just wasn't at the meeting where we all agreed that FEA should be regarded as assembly-equivalent, to be emitted by programs, and he's still trying to make it easy for humans to write. ;-)

justvanrossum: > I think Just wasn't at the meeting where we all agreed that FEA should be regarded as assembly-equivalent

TIL. Thanks, and agreed that's a good thing. Back to lurk mode!

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -310,6 +314,28 @@ the values of various table fields [§[9](#9)].
...
+
+Notes:
+ - A `<location_spec>` consists of one or more `<axis_tag>=<axis_location>` pairs, separated by commas 
+ - `<location_spec>:<value>` pairs are separated by spaces within the variable `<metric>`
+ - multiple `<location_spec>:<value>` pairs may be specified within a variable `<metric>`
+
+Example showing a simple single-axis variable scalar value. The value is -100 when Weight is 200, and -150 when Weight is 900:
+```
+<(wght=200:-100 wght=900:-150)>
+```
```
ctrlcctrlv, line 332: It's unclear what happens when Weight is neither 200 or 900.

ctrlcctrlv (changes requested):
Because conditionset's work on ranges, it's strange that §2.e.ii. metrics don't _seem_ to? §2.e.ii. is unclear.

docs/OpenTypeFeatureFileSpecification.md
josh-hadley: @simoncozens could you comment/clarify here?

docs/OpenTypeFeatureFileSpecification.md
josh-hadley: @moyogo `variation`s are very similar to `feature` definitions, but are stored differently in OT tables, and also have the possibly-NULL `conditionset` as part of their definition.

docs/OpenTypeFeatureFileSpecification.md
simoncozens: Now I've been actually *using* this, I can add more here:

* We should probably clarify that values are in userspace coordinates.
* We should probably clarify that a value corresponding to the default location (the default axis value on each axis) must be included in the variable scalar.
* We may want to provide shortcut syntax for this, where the default location is elided, I don't know; e.g. `<(-100 wght=900:-150)>`

In answer to Fred's question, we could add something like "The value of the variable scalar at locations other than the provided master locations will be interpolated according to OpenType Font Variation interpolation rules" if necessary, but to me it feels tautological.

simoncozens:
> Because conditionset's work on ranges, it's strange that §2.e.ii. metrics don't seem to? §2.e.ii. is unclear.

No, not at all.

Variable scalars are defining an interpolation model - describing how a value changes across the design space - by specifying masters values at particular locations in the design space. In this sense they are just like masters of a glyph outline: you wouldn't want a bold master to operate between wght=500 and wght=700 - it's necessarily a description of the outline at a particular point. So too is your "bold master" kern value: it operates at a point location, and its value is interpolated at non-master point locations.

Conditionsets are completely different, as they specify when something applies and when it doesn't.

simoncozens:
Just for interest, I wrote a utility called [ds2varlayout](https://github.com/simoncozens/ds2varlayout) which creates a single variable layout file (according to the syntax of this PR) from a designspace file and set of UFOs. It requires [this branch](https://github.com/simoncozens/fonttools/tree/fealib-variable-scalar) of fonttools.

Here is the result of running it on MutatorSans:

```fea
conditionset ConditionSet1 {
   wdth 0.0 328.0;
} ConditionSet1;

variation rvrn ConditionSet1 {
    sub I by I.narrow;
} rvrn;

conditionset ConditionSet2 {
   wdth 0.0 1000.0;
   wght 0.0 500.0;
} ConditionSet2;

variation rvrn ConditionSet2 {
    sub S by S.closed;
} rvrn;

@kern1.MMK_L_A = [A];
@kern2.MMK_R_A = [A];

lookup kern_ltr {
    lookupflag IgnoreMarks;
    pos A J (wdth=0,wght=0:0 wdth=0,wght=1000:-20 wdth=0,wght=700:0 wdth=1000,wght=700:0 wdth=569,wght=700:0);
    pos A O (wdth=0,wght=0:0 wdth=0,wght=1000:-30 wdth=0,wght=700:0 wdth=1000,wght=700:0 wdth=569,wght=700:0);
    # ...
    enum pos T @kern2.MMK_R_A (wdth=0,wght=0:-75 wdth=1000,wght=0:-215 wdth=1000,wght=1000:-150 wdth=0,wght=700:-75 wdth=1000,wght=700:-75 wdth=569,wght=700:-75);
    # ...
} kern_ltr;

feature kern {
    lookup kern_ltr;
} kern;
```

Hopefully it is easy to see how this enables variable-first building without having to merge layout tables between instance TTFs.

khaledhosny:
Can we have a syntax for “inline” condition sets as well? something like:
```fea
variation rvrn (wdth 0 328) {
    sub I by I.narrow;
} rvrn;

variation rvrn (wdth 0 1000, wght 0 500) {
    sub S by S.closed;
} rvrn;
```

Or would that be too much complication?

behdad:
I feel like the `variation` keyword is confusing. Why not just stick with `feature` and add conditionset?

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -1267,6 +1292,44 @@ rules, the implementation sorts the rules to avoid conflict; for example, the
...
+For example, to define a "heavy" condition, for the 'wght' axis between 700-900:
+```fea
+conditionset heavy {
+    wght 700 900;
+} heavy;
+```
+
+<a name="4.e.i.2"></a>
+##### 4.e.i.2 `variation` (FeatureVariation)
+A `variation` is akin to a regular `feature` definition, with an added `conditionset` specifier to indicate that the feature should be active when the conditions of the conditionset are met. The syntax is:
```
simoncozens, line 1320: OT gives us feature table *substitutions* - in the sense that one feature table is swapped out and another is swapped in. So if you have:

```
feature test {
   sub a by b;
} test;

variation test heavy {
   sub d by e;
} test;
```

then if the heavy condition is met, the variation gets swapped in, the original feature gets swapped out, and *only* the `d->e` substitution will happen, as I understand it. 

If this is the intent, then this should be clearer in the text here. I suggest replacing "to indicate that the feature should active" (which is kind of wooly anyway - "active"?) with "to indicate that the feature should be replaced by the rules contained in the `variation` statement".

simoncozens:
I believe I now have a complete implementation of this syntax in https://github.com/fonttools/fonttools/pull/2228

ctrlcctrlv:
> Just for interest, I wrote a utility called [ds2varlayout](https://github.com/simoncozens/ds2varlayout) which creates a single variable layout file (according to the syntax of this PR) from a designspace file and set of UFOs. It requires [this branch](https://github.com/simoncozens/fonttools/tree/fealib-variable-scalar) of fonttools.
> 
> Here is the result of running it on MutatorSans:
> 
> ```opentype
> conditionset ConditionSet1 {
>    wdth 0.0 328.0;
> } ConditionSet1;
> 
> variation rvrn ConditionSet1 {
>     sub I by I.narrow;
> } rvrn;
> 
> conditionset ConditionSet2 {
>    wdth 0.0 1000.0;
>    wght 0.0 500.0;
> } ConditionSet2;
> 
> variation rvrn ConditionSet2 {
>     sub S by S.closed;
> } rvrn;
> 
> @kern1.MMK_L_A = [A];
> @kern2.MMK_R_A = [A];
> 
> lookup kern_ltr {
>     lookupflag IgnoreMarks;
>     pos A J (wdth=0,wght=0:0 wdth=0,wght=1000:-20 wdth=0,wght=700:0 wdth=1000,wght=700:0 wdth=569,wght=700:0);
>     pos A O (wdth=0,wght=0:0 wdth=0,wght=1000:-30 wdth=0,wght=700:0 wdth=1000,wght=700:0 wdth=569,wght=700:0);
>     # ...
>     enum pos T @kern2.MMK_R_A (wdth=0,wght=0:-75 wdth=1000,wght=0:-215 wdth=1000,wght=1000:-150 wdth=0,wght=700:-75 wdth=1000,wght=700:-75 wdth=569,wght=700:-75);
>     # ...
> } kern_ltr;
> 
> feature kern {
>     lookup kern_ltr;
> } kern;
> ```
> 
> Hopefully it is easy to see how this enables variable-first building without having to merge layout tables between instance TTFs.

I notice decimal numbers in the quoted file. As far as I'm aware there are no decimal numbers in FEA grammar. Was this intentional?

simoncozens:
No. It's fixed now.

anthrotype:
but we _do_ want to be able to use floats for locations (fvar has `Fixed` 16.16 for user-space min/default/max, and FeatureVariations Condition has F2Dot14 for normalized internal min/max coords)

simoncozens:
The positions in the feature file are un-normalized, userspace coordinates. They should be normalised by the feature compiler when generating GDEF variable scalars and FeatureVariations conditions.

anthrotype:
> The positions in the feature file are un-normalized, userspace coordinates.

I know. But even user-space coordinates in VFs can be floats (fvar Fixed) so feature file should allow that too.

twardoch:
I intuitively like the `variation` keyword. And I think we have agreement and movement on this issue, which I'm very glad about, given that it's been dormant for five years since I raised it :)

behdad:
I know I wasn't at the meeting that this was discussed. But I've been thinking about the proposed syntax and I *highly* encourage you to reconsider a couple of things.

Simon wrote:

> OT gives us feature table _substitutions_ - in the sense that one feature table is swapped out and another is swapped in. So if you have:
> 
> ```
> feature test {
>    sub a by b;
> } test;
> 
> variation test heavy {
>    sub d by e;
> } test;
> ```
> 
> then if the heavy condition is met, the variation gets swapped in, the original feature gets swapped out, and _only_ the `d->e` substitution will happen, as I understand it.
> 
> If this is the intent, then this should be clearer in the text here. I suggest replacing "to indicate that the feature should be active" (which is kind of wooly anyway - "active"?) with "to indicate that the feature should be replaced by the rules contained in the `variation` statement".

In fact, I think this is extremely cumbersome to use. Since the "varied feature" needs to list everything that applies at that point.

Instead, I'm of the belief that the FEA syntax should allow conditioning on the lookup / rule level within each feature, and let the compiler figure out how to translate that to OT1.8.

So I suggest an alternative that is more like code:
```
feature test {
   sub a by b;
   sub d by e if heavy;
} test;
```

docs/OpenTypeFeatureFileSpecification.md
behdad: > In fact, I think this is extremely cumbersome to use. Since the "varied feature" needs to list everything that applies at that point.

In other words, the current design would require would to call `fontTools.varLib.featureVars`-like code to resolve rules that apply at every region to generate the .fea file. Whereas in my proposal, the designer intentions are encoded directly and resolved when compiling the font.

cc @justvanrossum

behdad:
> In fact, I think this is extremely cumbersome to use. Since the "varied feature" needs to list everything that applies at that point.

In other words, the current design would require would to call fontTools.varLib.featureVars-like code to resolve rules that apply at every region to generate the .fea file. Whereas in my proposal, the designer intentions are encoded directly and resolved when compiling the font.

cc @justvanrossum

Lorp:
Can we please clarify why locations must be specified in user-space? If “we all agreed that FEA should be regarded as assembly-equivalent”, then isn’t it a little bit odd? I’m not against allowing user-space, but this does represent a major proliferation of it in font sources, so seems to me to require justification to overcome possible fragility. I’m thinking in particular of the common use case of changing source locations late in the design process.

How about allowing syntax using `default`, `min` and `max` keywords, representing normalized 0, -1 and 1, as an alternative to user-space? Compare the following example, which expresses the same thing in user-space, then in normalized space. (Axes: wght 100/**400**/800, wdth 75/**100**/100)

```
# user-space, needs every location updated if source locations change

pos A J (
    wght=400,wdth=100:0
    wght=400,wdth=75:-20
    wght=100,wdth=100:-20
    wght=100,wdth=75:-20
    wght=800,wdth=100:-10
    wght=800,wdth=75:-10
);

# normalized, allows source locations to be changed without needing updates here

pos A J (
    wght=default,wdth=default:0
    wght=default,wdth=min:-20
    wght=min,wdth=default:-20
    wght=min,wdth=min:-20
    wght=max,wdth=default:-10
    wght=max,wdth=min:-10
);
```

Syntax is also needed for intermediate locations, for which I suggest `0.5min`, `0.812max` etc. Again, such specifications are robust in design terms when source locations change.

twardoch:
I think most people will find it challenging that locations in FEA are not in designer space. But user space is still approachable to them. Normalized space is not, I think. It's two times removed from how people usually work. 

Also, normalized space locations change dramatically when you relocate the neutral master. I think user space is the most sensible solution, the best compromise. 

Also, is normalized space, what units? The theoretical –1 to 1 with an unpredictable decimal precision, or the real –16k to 16k, or something else?

Lorp:
To clarify, I’m proposing not a replacement to user-space syntax but an alternative syntax that can be used if it’s better for one’s workflow. I’ve repeated the alternative syntax below with implicit defaults to show potential brevity.

```
pos A J (
    0
    wdth=min:-20
    wght=min:-20
    wght=min,wdth=min:-20
    wght=max:-10
    wght=max,wdth=min:-10
);
```

Is there a spec of how arbitrary user-space locations get translated into variation regions?

simoncozens:
> Is there a spec of how arbitrary user-space locations get translated into variation regions?

I don't mean to be facetious, but this is almost certainly the wrong question. Variation regions for the calculation of variable scalar supports is purely a matter of internal representation within the GDEF/gvar tables. Nothing good will come of exposing this to the user.

Lorp:
Let’s say we have the same wght 100/**400**/800, wdth 75/**100**/100 font described above. Currently I don’t know whether the following feature files are valid, nor how they are supposed to compile if they are. Apologies if I’m being dumb.

```
# default specified, one location specified that is intermediate on both axes
pos A J (
    wght=400,wdth=100:0
    wght=600,wdth=85:0
);
```

```
# default unspecified, one location specified that is intermediate on both axes
pos A J (
    wght=600,wdth=85:0
);
```

```
# default unspecified, max on one axis specified
pos A J (
    wght=800:0
);
```

simoncozens:
> Let’s say we have the same wght 100/**400**/800, wdth 75/**100**/100 font described above. Currently I don’t know whether the following feature files are valid, nor how they are supposed to compile if they are. Apologies if I’m being dumb.
> 
> ```
> # default specified, one location specified that is intermediate on both axes
> pos A J (
>     wght=400,wdth=100:0
>     wght=600,wdth=85:0
> );
> ```
> 
>  ```
> # default unspecified, one location specified that is intermediate on both axes
> pos A J (
>     wght=600,wdth=85:0
> );
> ```
> 
>  ```
> # default unspecified, max on one axis specified
> pos A J (
>     wght=800:0
> );
> ```

The first one is valid. The other two are not. You need to specify the value for the default position. (Again, this is *exactly* like outline interpolation.) You can also try compiling these using the fontTools branch mentioned above.

moyogo:
FYI: @ErwinDenissen implemented @simoncozens proposal in Font Creator 14, see [the announcement on TypeDrawers](https://typedrawers.com/discussion/4111/new-fontcreator-14-supports-variable-fonts).

simoncozens:
That means there are now two independent implementations of this! Nice!

@josh-hadley, what’s needed to drive this forward? (I have a horrible feeling you’re going to say something about makeotf...)

behdad:
Problem with "designspace coordinates" is that, such a concept is undefined by OpenType and FEA file, and only defined by the Designspace file and varLib.

justvanrossum:
So am I correct to summarize the .fea design choices available here as:

- userspace coordinates
- normalized coordinates pre avar
- normalized coordinates post avar

?

simoncozens:
I think putting normalised (-1.0<=>1.0) coordinates into the FEA file would be really unpleasant for the user. I know I’ve previously said that humans shouldn’t write FEA, but they probably should be able to understand it! But of course they will end up as normalised values in the GDEF variation store.

From my perspective, the choice is coming down to being between userspace values as taken from the fvar table (wght=400 in the Nunito example), which is how we’ve interpreted things so far; or designspace values (wght=42) plus some additional syntax allowing the user to provide a mapping which tells the compiler how to turn that into a normalised value.

behdad:
> So am I correct to summarize the .fea design choices available here as:
> 
> * userspace coordinates
> * normalized coordinates pre avar
> * normalized coordinates post avar
> 
> ?

What I suggest is that, but in Designspace file and Fea file, we allow specifying the coordinate space. And allow the following coordinate spaces:

- userspace
- normalized userspace
- designspace
- normalized designspace
- integer normalized designspace in F2DOT14

If designspace coordinate system is used in .fea file, then it either references an externally defined coordinate space, or a mapping ala one from Designspace should be specified in .fea file in a new syntax to be defined.

So yes, we need avar and fvar syntax as well.

josh-hadley:
Folks -- thanks for the renewed discussion on this. It seems like there's still sufficient disagreement on some finer (and major) points to warrant a[nother] meeting or maybe workshop of sorts of stakeholders to try to get agreement on something that we can solidify in FEA syntax. If people are open to that, I can work on scheduling something.

josh-hadley:
Hi folks, I just wanted to let people know that we are still actively working on updates around FEA variable syntax.

madig:
I, uh, how do user-space locations in a feature file interact with mapping-laced user locations in a Designspace? I found a case where a mapping in a Designspace file like so:

```xml
  <axes>
    <axis tag="wght" name="Weight" minimum="100" maximum="900" default="100">
      <map input="100" output="22"/>
      <map input="600" output="115"/>
      <map input="900" output="185"/>
    </axis>
  </axes>
  <rules processing="last">
    <rule name="conditional_sub">
      <conditionset>
        <condition name="Weight" minimum="116" maximum="185"/>
      </conditionset>
      <sub name="e" with="a"/>
    </rule>
  </rules>
```

means that "e" is substituted by "a" at wght=606 and upwards (116 in design coordinates is a tad over 600 in user coordinates), whereas with a feature file snippet like this:

```
lookup conditional_sub {
    sub e by a;
} conditional_sub;

conditionset test {
    wght 606 900;
} test;

variation rlig test {
    lookup conditional_sub;
} rlig;
```

it is substituted at wght=644 and upwards?

anthrotype:
hey @josh-hadley, following the activity on the fonttools end, I wonder if you're still working on this FEA spec update?

skef:
@anthrotype Yes, the Adobe font folks are actively synthesizing the feedback on this PR and other ideas into a new proposal and draft of relevant portions of the AFDKO. Sometimes that work has to take a back seat to other things that come up but we're working on it.

davelab6:
@josh-hadley would an online workshop day or two be helpful? I'm interested to organize one for following up on the atypi panel. 

Will anyone here attend typographics/typelab in NYC in June or typecon in Portland Oregon in August?

anthrotype:
@schriftgestalt might be interested in implementing this in Glyphs.app, now that fontTools feaLib and fea-rs added support for this proposed variable FEA syntax. Can we do something to unblock this?
I see there are merge conflicts, maybe @simoncozens can try rebasing his PR.

docs/OpenTypeFeatureFileSpecification.md
anthrotype: I missed this important point when I first reviewed this. The way we implemented int feaLib, by using featureVars.addFeartureVariationsRaw, means that the the new Feature table that replaces the default one when the conditions are met simply extends the existing feature's lookup list indices, but does not replaces it:

https://github.com/fonttools/fonttools/blob/7412268cbe2a36f588e2d06b45faf35a3b803633/Lib/fontTools/varLib/featureVars.py#L429-L436

Maybe we should try to support both use-cases somehow, making this explicit at the syntax level.

skef:
@anthrotype Although we understand this is taking a long time, due to a lot of reasons including resources needed to keep up with the ongoing OpenType spec work, 

1. Adobe is still developing the VF-first feature file specification and AFDKO implemenation
2. This PR does not represent the current thinking, and there are some aspects of the proposal (e.g. the low-level syntax for the substitution portion) that we would not recommend).

rsheeter:
@skef when do you expect to be able to share more? Even a preliminary draft would be helpful, otherwise the implementations will (continue to) drift.

schriftgestalt:
Just to mention that GlyphsApp supports it’s own version of this. By the time we implemented this, there was no discussion or active proposal around.
https://handbook.glyphsapp.com/layout/variable-gpos/
https://handbook.glyphsapp.com/layout/conditional-feature-code/

(There are some more simple improvements to the fea syntax that we implemented and like to propose when it seems useful).

skef:
I forgot to add a comment here until now: This PR has been partly superseded by a newer proposal now under review (until January 15th 2024) at https://github.com/adobe-type-tools/feature_file_change_review . The implementation replaces the variable value portion of this work. There will be a second phase, probably in the first half of 2024, to address the remaining Feature Variations cases. We expect that proposal to be somewhat different from the draft here as well.

I'll close this PR once we have a draft of the phase 2 proposal.

ErwinDenissen:
I hope all font software vendors adopt this new feature syntax as soon as possible, as it will significantly enhance interchangeability between source font file formats.

P.S. In the link you provided, you refer to January 15th **2025** and the first half of **2025**, which I assume is what you meant here as well.

skef:
Ah, yes, thank you. Edited.

PR #1353, opened 2021-04-30
uforead, ufowrite, checkoutlinesufo fixes for reading & writing CID-keyed fonts

kaydeearts:
This PR adds support for CID-keyed font info previously lost in `tx` & `checkoutlinesufo` during reading and writing.


josh-hadley (approved):
LGTM. There is 1 commented-out line that can be deleted.

josh-hadley (approved):
🎉

PR #1362, opened 2021-05-17
Clarify the use of hyphens in glyph names

### LINKED ISSUE #970: [spec] Outdated statements about glyph classes
From [2.g.i](https://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#2gi-ranges):

> A range of glyphs is denoted by a hyphen:
> ```
> [<firstGlyph> - <lastGlyph>]
> ```
> Spaces around the hyphen are not required since hyphens are not permitted in feature file glyph names. For example:
> ```
> [\0-\31]
> [A-Z]
> ```

But I believe this is no longer the case since hyphen in glyph names are permitted now.

iterumllc:
## Description

Having hyphens at the end of "development" glyph names makes handling ranges correctly in the grammar significantly trickier; the current implementation doesn't attempt to. As there doesn't seem to be a need for them in that location this PR updates the docs to clarify that they can't end a name.

Perhaps more controversially it also states that the special characters can't be used at the start of a name. The spec was ambiguous about this but the `featgram.g` implementation doesn't allow them at the beginning of a name. I agree with that decision—too dicey to allow them there. 

One additional question is whether we should go farther and restrict the use of more or all of the special characters at the end of a name. The main reason for doing that would be to reserve potential glyph operators for future use, which could be handled the way hyphens are handled now. I leave that up for discussion. 

## Checklist:

Closes #970

### DIFF
--- a/docs/OpenTypeFeatureFileSpecification.md
+++ b/docs/OpenTypeFeatureFileSpecification.md
@@ -10,8 +10,8 @@ Copyright 2015-2021 Adobe. All Rights Reserved. This software is licensed as
 OpenSource, under the Apache License, Version 2.0. This license is available at:
 http://opensource.org/licenses/Apache-2.0.
 
-Document version 1.26
-Last updated 7 June 2021
+Document version 1.26.1
+Last updated 1 Aug 2023
 
 **Caution: Portions of the syntax unimplemented by Adobe are subject to change.**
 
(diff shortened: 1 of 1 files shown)


skef:
I'll update this and assign @kaydeearts soon

skef:
We should add this hyphen restriction to the spec. I doubt it will ruffle too many feathers. I'll take a todo on it.

PR #1360, opened 2021-05-14
Switch to CMake-based builds

iterumllc:
## Description

Some of the likely further development of this repository would benefit from simplification of the C code build system. This PR removes the old files and replaces them with a CMake-driven build started off by the import of `scikit-build` into `setup.py`.

This method of building non-python code within python packages is now supposed to be widely supported on Linux, Mac OS X, and more recent versions of Visual Studio. I have only tested on Linux, however, so one active role of this PR is to see how the build process goes on the other platforms and make any needed adjustments.

Some source directories are moved to more standard locations. The documentation of manual builds in `doc/FDK_Build_Notes.md` has been revised. Various workflow files were slightly changed. 

What this PR does *not* change:

1. It has no new C or Python code.
2. It attempts to retain the previous compilation defines.
3. ~~The release builds are still compiled without optimization and the executables are still not stripped.~~
4. The executables are still supplied directly as "scripts", although we should consider putting them in platform-specific data directories and using entrypoint-based wrappers as in https://github.com/scikit-build/ninja-python-distributions/tree/master/ninja , as this would be an overall cleaner and more accepted solution. 

Added: As of 4e7241ba0b8203c7de1c1280c60abfb220e3fd7b the default build type is now "RelWithDebInfo", which combines optimization and putting debugging symbols into the final binary. While the latter takes up some extra space it can be useful for users who want to step through the programs in a debugger. 

## Checklist:

- [x] I have followed the [Contribution Guidelines](https://github.com/adobe-type-tools/afdko/blob/develop/CONTRIBUTING.md)
- [ ] I have added **test code and data** to prove that my code functions correctly
- [x] I have verified that new and existing tests pass locally with my changes
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation

Closes #1124

Issue #114, opened 2016-03-19
[spec] Includes not ending in semicolon


khaledhosny:
It seems that `makeotf` allows include statements not ending with semicolons, e.g.:

```
include (markclasses.fea)
```

But the spec does not document this, so two questions:
- Is this a bug or an undocumented feature?
- What happens if other feature code is present in the same line, e.g.

```
include (markclasses.fea) feature foo {} foo;
```

readroberts:
The spec does show that an include statement should be terminated by a semi-colon. I would like to keep it this way. Unfortunately, there are a lot of old source files in the AdobeType library whcihdoes not terminate with a semi-colon. I will change makeotf to issue a warning, to discourage this practice, but not a fatal error. 
If someone really wants to add more definitions after a feature file include statement that is missing a semi-colon, I would issue a warning, but otherwise behave as if there were a semi-colon after the include statement.

khaledhosny:
Thanks for explanation.

frankrolf:
In my opinion, it makes sense for `include (file.fea)` statements not to end in a semicolon, because the contents of `file.fea` most likely does.

Consider the contents of the following `file.fea`:
```
feature smcp { 
sub [a-z] by [a.sc-z.sc];
} smcp;
```

The statement `include (file.fea);` would basically create:
```
feature smcp { 
sub [a-z] by [a.sc-z.sc];
} smcp;;
```
(note last semicolon)

khaledhosny:
Unless the parser “consumes” the semicolon after parsing the include statement, which is the logical thing to do.

readroberts:
Khaled is correct; the parser consumes the terminating semicolon, if it is present.

miguelsousa:
The warning was added in 49527cd0cb65982e0151571c2df631389ba900ee

PR #1367, opened 2021-05-31
Port of Feature File Parser to Antlr 4

### LINKED ISSUE #83: [makeotfexe] hotconv featgram.c Xcode compiler: many errors from unnecessary parens in featgram.c
The -Wparentheses-equality warning is useful for more than just extra parens, and I'd like to keep it. Solved the problem by fixing the source code in  featgram.c. A short -term solution: I replaced all occurrences of "(if|while) (\s+(LA(1)==([^)]+))\s*)" with " \1 (LA(1)==\2)". Short-term because this will have to be repeated every time someone regenerates the featgram.c file by running the ANTLR parser. Needs to eventually be fixed in the ANTLR parser itself.

### LINKED ISSUE #548: revisit how we use ANTLR
Currently we have an old copy of ANTLR in the makeotf source tree. I think we should switch over to using the latest version, perhaps pulling it down as a git submodule or using a download from their homepage, rather than keeping a copy in our source tree.

Here's ANTLR on GitHub:
https://github.com/antlr

Here's the main ANTLR homepage:
http://www.antlr.org

### LINKED ISSUE #1124: Better build system
Currently there are 3 build systems for the C tools (one for each of macOS, Windows, and Linux, and there is even a second Linuix one that I think not actually used).

Adding a single source file requires editing 6 files and using XCode and Visual Studio and verifying the build on 3 platforms.

I think the C tools can benefit from using a build system that consolidates this. I think either CMake or meson might be a good idea, I have a bit of experience with meson and it can be installed from PyPI but it sometimes feels a bit immature and every once in a while a new version breaks the build.

```diff
diff --git a/c/makeotf/makeotf_lib/build/hotconv/linux/debug/Makefile b/c/makeotf/makeotf_lib/build/hotconv/linux/debug/Makefile
index 8b1b0dd3..def94e4e 100644
--- a/c/makeotf/makeotf_lib/build/hotconv/linux/debug/Makefile
+++ b/c/makeotf/makeotf_lib/build/hotconv/linux/debug/Makefile
@@ -32,12 +32,13 @@ LIB_SRCS = \
 	$(SRC_DIR)/map.c \
 	$(SRC_DIR)/otl.c \
 	$(SRC_DIR)/sfnt.c \
+	$(SRC_DIR)/STAT.c \
 	$(SRC_DIR)/featerr.c \
 	$(SRC_DIR)/featgram.c \
 	$(SRC_DIR)/featscan.c
 LIB_OBJS = hot.o BASE.o CFF_.o cmap.o fvar.o GPOS.o GSUB.o head.o hhea.o \
 	hmtx.o maxp.o MMFX.o MMSD.o name.o OS_2.o post.o vhea.o vmtx.o VORG.o \
-	anon.o map.o otl.o sfnt.o featerr.o featgram.o featscan.o
+	anon.o map.o otl.o sfnt.o STAT.o featerr.o featgram.o featscan.o
 LIB_TARGET = $(LIB_DIR)/hotconv.a
 
 # Build targets
@@ -119,6 +120,9 @@ otl.o:
 sfnt.o:
 	$(CC) $(CFLAGS) -c $(SRC_DIR)/sfnt.c -o $@
 
+STAT.o:
+	$(CC) $(CFLAGS) -c $(SRC_DIR)/STAT.c -o $@
+
 featerr.o:
 	$(CC) $(CFLAGS) -c $(SRC_DIR)/featerr.c -o $@
 
diff --git a/c/makeotf/makeotf_lib/build/hotconv/linux/gcc/debug/Makefile b/c/makeotf/makeotf_lib/build/hotconv/linux/gcc/debug/Makefile
index daae26ca..96d5fbfe 100755
--- a/c/makeotf/makeotf_lib/build/hotconv/linux/gcc/debug/Makefile
+++ b/c/makeotf/makeotf_lib/build/hotconv/linux/gcc/debug/Makefile
@@ -47,6 +47,7 @@ LIB_SRCS = \
 	$(SRCDIR)/otl.c \
 	$(SRCDIR)/post.c \
 	$(SRCDIR)/sfnt.c \
+	$(SRCDIR)/STAT.c \
 	$(SRCDIR)/vhea.c \
 	$(SRCDIR)/vmtx.c \
 	$(SRCDIR)/VORG.c
@@ -76,6 +77,7 @@ LIB_OBJS = \
 	$(OBJECT_DIR)/otl.o \
 	$(OBJECT_DIR)/post.o \
 	$(OBJECT_DIR)/sfnt.o \
+	$(OBJECT_DIR)/STAT.o \
 	$(OBJECT_DIR)/vhea.o \
 	$(OBJECT_DIR)/vmtx.o \
 	$(OBJECT_DIR)/VORG.o
@@ -157,6 +159,9 @@ post.o:
 sfnt.o:
 	$(CC) $(CFLAGS) -c $(SRC_DIR)/sfnt.c -o $@
 
+STAT.o:
+	$(CC) $(CFLAGS) -c $(SRC_DIR)/STAT.c -o $@
+
 vhea.o:
 	$(CC) $(CFLAGS) -c $(SRC_DIR)/vhea.c -o $@
 
diff --git a/c/makeotf/makeotf_lib/build/hotconv/linux/gcc/release/Makefile b/c/makeotf/makeotf_lib/build/hotconv/linux/gcc/release/Makefile
index 4ad3d007..4819bcdc 100755
--- a/c/makeotf/makeotf_lib/build/hotconv/linux/gcc/release/Makefile
+++ b/c/makeotf/makeotf_lib/build/hotconv/linux/gcc/release/Makefile
@@ -46,6 +46,7 @@ LIB_SRCS = \
 	$(SRCDIR)/otl.c \
 	$(SRCDIR)/post.c \
 	$(SRCDIR)/sfnt.c \
+	$(SRCDIR)/STAT.c \
 	$(SRCDIR)/vhea.c \
 	$(SRCDIR)/vmtx.c \
 	$(SRCDIR)/VORG.c
@@ -75,6 +76,7 @@ LIB_OBJS = \
 	$(OBJECT_DIR)/otl.o \
 	$(OBJECT_DIR)/post.o \
 	$(OBJECT_DIR)/sfnt.o \
+	$(OBJECT_DIR)/STAT.o \
 	$(OBJECT_DIR)/vhea.o \
 	$(OBJECT_DIR)/vmtx.o \
 	$(OBJECT_DIR)/VORG.o
@@ -156,6 +158,9 @@ post.o:
 sfnt.o:
 	$(CC) $(CFLAGS) -c $(SRC_DIR)/sfnt.c -o $@
 
+STAT.o:
+	$(CC) $(CFLAGS) -c $(SRC_DIR)/STAT.c -o $@
+
 vhea.o:
 	$(CC) $(CFLAGS) -c $(SRC_DIR)/vhea.c -o $@
 
diff --git a/c/makeotf/makeotf_lib/build/hotconv/linux/release/Makefile b/c/makeotf/makeotf_lib/build/hotconv/linux/release/Makefile
index a461fdcf..d459d899 100644
--- a/c/makeotf/makeotf_lib/build/hotconv/linux/release/Makefile
+++ b/c/makeotf/makeotf_lib/build/hotconv/linux/release/Makefile
@@ -32,12 +32,13 @@ LIB_SRCS = \
 	$(SRC_DIR)/map.c \
 	$(SRC_DIR)/otl.c \
 	$(SRC_DIR)/sfnt.c \
+	$(SRC_DIR)/STAT.c \
 	$(SRC_DIR)/featerr.c \
 	$(SRC_DIR)/featgram.c \
 	$(SRC_DIR)/featscan.c
 LIB_OBJS = hot.o BASE.o CFF_.o cmap.o fvar.o GPOS.o GSUB.o head.o hhea.o \
 	hmtx.o maxp.o MMFX.o MMSD.o name.o OS_2.o post.o vhea.o vmtx.o VORG.o \
-	anon.o map.o otl.o sfnt.o featerr.o featgram.o featscan.o
+	anon.o map.o otl.o sfnt.o STAT.o featerr.o featgram.o featscan.o
 LIB_TARGET = $(LIB_DIR)/hotconv.a
 
 # Build targets
@@ -119,6 +120,9 @@ otl.o:
 sfnt.o:
 	$(CC) $(CFLAGS) -c $(SRC_DIR)/sfnt.c -o $@
 
+STAT.o:
+	$(CC) $(CFLAGS) -c $(SRC_DIR)/STAT.c -o $@
+
 featerr.o:
 	$(CC) $(CFLAGS) -c $(SRC_DIR)/featerr.c -o $@
 
diff --git a/c/makeotf/makeotf_lib/build/hotconv/osx/xcode/hotconv.xcodeproj/project.pbxproj b/c/makeotf/makeotf_lib/build/hotconv/osx/xcode/hotconv.xcodeproj/project.pbxproj
index 20845ca0..924da5fe 100644
--- a/c/makeotf/makeotf_lib/build/hotconv/osx/xcode/hotconv.xcodeproj/project.pbxproj
+++ b/c/makeotf/makeotf_lib/build/hotconv/osx/xcode/hotconv.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		29B88DDE244FA4DD006F259E /* STAT.h in Headers */ = {isa = PBXBuildFile; fileRef = 29B88DDC244FA4DC006F259E /* STAT.h */; };
+		29B88DE1244FA500006F259E /* STAT.c in Sources */ = {isa = PBXBuildFile; fileRef = 29B88DE0244FA500006F259E /* STAT.c */; };
 		BD1F638E0E26B32700A4E4C9 /* GDEF.h in Headers */ = {isa = PBXBuildFile; fileRef = BD1F638D0E26B32700A4E4C9 /* GDEF.h */; };
 		BD1F63900E26B33800A4E4C9 /* GDEF.c in Sources */ = {isa = PBXBuildFile; fileRef = BD1F638F0E26B33800A4E4C9 /* GDEF.c */; };
 		BD6159A50D2DC53300C1E0CA /* post.c in Sources */ = {isa = PBXBuildFile; fileRef = BD61598E0D2DC53300C1E0CA /* post.c */; };
@@ -76,6 +78,8 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXFileReference section */
+		29B88DDC244FA4DC006F259E /* STAT.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = STAT.h; path = ../../../../source/hotconv/STAT.h; sourceTree = "<group>"; };
+		29B88DE0244FA500006F259E /* STAT.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; name = STAT.c; path = ../../../../source/hotconv/STAT.c; sourceTree = "<group>"; };
 		BD1F638D0E26B32700A4E4C9 /* GDEF.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = GDEF.h; path = ../../../../source/hotconv/GDEF.h; sourceTree = SOURCE_ROOT; };
 		BD1F638F0E26B33800A4E4C9 /* GDEF.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; name = GDEF.c; path = ../../../../source/hotconv/GDEF.c; sourceTree = SOURCE_ROOT; };
 		BD61598E0D2DC53300C1E0CA /* post.c */ = {isa = PBXFileReference; explicitFileType = sourcecode.c.c; fileEncoding = 4; name = post.c; path = ../../../../source/hotconv/post.c; sourceTree = SOURCE_ROOT; };
@@ -173,6 +177,7 @@
 		08FB7795FE84155DC02AAC07 /* Source */ = {
 			isa = PBXGroup;
 			children = (
+				29B88DE0244FA500006F259E /* STAT.c */,
 				BD6159BC0D2DC54200C1E0CA /* CFF_.c */,
 				BD6159BE0D2DC54200C1E0CA /* BASE.c */,
 				BD6159BF0D2DC54200C1E0CA /* anon.c */,
@@ -224,6 +229,7 @@
 		BD6159C40D2DC55D00C1E0CA /* Headers */ = {
 			isa = PBXGroup;
 			children = (
+				29B88DDC244FA4DC006F259E /* STAT.h */,
 				BD9DCE190DB540DE00192F3D /* hotconv.h */,
 				BDE662050D329ED5005FE583 /* feat.c */,
 				BD6159C50D2DC56500C1E0CA /* post.h */,
@@ -281,6 +287,7 @@
 			isa = PBXHeadersBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				29B88DDE244FA4DD006F259E /* STAT.h in Headers */,
 				BD6159E90D2DC56500C1E0CA /* post.h in Headers */,
 				BD6159EA0D2DC56500C1E0CA /* otl.h in Headers */,
 				BD6159EC0D2DC56500C1E0CA /* map.h in Headers */,
@@ -387,6 +394,7 @@
 				BD6159B00D2DC53300C1E0CA /* hmtx.c in Sources */,
 				BD6159B10D2DC53300C1E0CA /* GSUB.c in Sources */,
 				BD6159B20D2DC53300C1E0CA /* MMSD.c in Sources */,
+				29B88DE1244FA500006F259E /* STAT.c in Sources */,
 				BD6159B30D2DC53300C1E0CA /* sfnt.c in Sources */,
 				BD6159B40D2DC53300C1E0CA /* OS_2.c in Sources */,
 				BD6159B50D2DC53300C1E0CA /* name.c in Sources */,
diff --git a/c/makeotf/makeotf_lib/build/hotconv/win/visualstudio/hotconv.vcxproj b/c/makeotf/makeotf_lib/build/hotconv/win/visualstudio/hotconv.vcxproj
index 1c057170..d6112058 100644
--- a/c/makeotf/makeotf_lib/build/hotconv/win/visualstudio/hotconv.vcxproj
+++ b/c/makeotf/makeotf_lib/build/hotconv/win/visualstudio/hotconv.vcxproj
@@ -160,6 +160,7 @@
     <ClCompile Include="..\..\..\..\source\hotconv\otl.c" />
     <ClCompile Include="..\..\..\..\source\hotconv\post.c" />
     <ClCompile Include="..\..\..\..\source\hotconv\sfnt.c" />
+    <ClCompile Include="..\..\..\..\source\hotconv\STAT.c" />
     <ClCompile Include="..\..\..\..\source\hotconv\vhea.c" />
     <ClCompile Include="..\..\..\..\source\hotconv\vmtx.c" />
     <ClCompile Include="..\..\..\..\source\hotconv\VORG.c" />
```

iterumllc:
## Description

This is primarily a port of the feature file parser to Antlr 4.9. As Antlr 4 does not have a C backend the port is written in C++, and that change requires a number of other changes diminishing with distance from the `hotconv` library. In the library itself some of the code needed to be restructured so that include files can be shared. At the application level the "main" function is now a C++ stub (as compiling main as C++ is the traditional way of ensuring that statics are initialized, etc.) Further out under the `shared` (formerly `public`) directory there is some `const` change leakage.

It also changes the build to be CMake-based (this PR supersedes #1360). The build should now work in XCode and Visual Studio. The tests have also been shimmed into cmake/ctest (at a coarse grain) so that they can be run prior to installation without fiddling. 

The PR is partly self-documenting in that changes to files in the `docs` explain both how to use the new build system and a bit about how the new Feature File parse tree translator works. The specification has also been updated to reflect some clarifications and changes in light of the new parser.

All tests pass. There is one new (aalt-related) test and some changes to tests. Most of the changes just sync up with the new console message format and/or library and program version numbers. `eii-1.bad.fea` was edited to be genuinely beyond the lower bound of `int16_t`.  `expected_output/fealib/GPOS_3.ttx` had a bad `AnchorPoint` value due to a bug in the old `feat.c` code.

The new system has also been tested against a number of full-fledged fonts by comparing the `ttx` output of files prepared with the old and new versions of `makeotfexe`. Even so, given the scope of the feature file semantics it is unlikely we have full test coverage of all relevant cases.

Closes #1124 
Closes #548 
Closes #83 
 
## Checklist:


iterumllc:
One of the goals of this port was to keep the lexer and parser grammars as free of implementation as possible, to enable other projects to use them for parsers in other target languages. Given some small changes to the spec (e.g. slightly narrowing the set of valid tags) the only part of the spec that required some implementation was `anon` blocks (to match the end tag up with the start tag while lexing, see also #1352).

Accordingly, the lexer is broken up into `FeatLexerBase.g4` (with almost all of the tokens, and free of any implementation) and `FeatLexer.g4` (which "imports" `FeatLexerBase` and adds anon-block support). `FeatParser.g4` uses the token set generated from `FeatLexer` and contains no implementation. Therefore any project wanting to use these grammars can port `FeatLexer.g4` to their target language and then just use the other files unmodified.

iterumllc:
One other aspect of this: I've tried to replace immediate FATALs with ERRORs where possible to increase the information available from a given run. Sometimes this requires adding a bit of fake data. In any case this increases the chances of hitting a crash (segfault, etc.) on error paths. If someone hits one of these they are easy enough to fix, either by adding more fake data or (easier) by putting the FATAL error back.

iterumllc:
The lgtm failure is a repeated github timeout (downloading utfccp, which is brought in by the Antlr 4 runtime). I've heard at least one other tester report this, but its strange coming from the github build infrastructure itself. 

If we absolutely have to copy everything into the repo and restructure the build around that we can but it seems beset to avoid if possible.

iterumllc:
I mitigated the timeout by running

```
git config --global url.https://github.com/.insteadOf git://github.com/
```

in the workflow before building.

tests/makeotfexe_data/input/bug993/feat.fea
```diff
@@ -1,5 +1,5 @@
 markClass [acute grave] <anchor 150 -10> @TOP_MARKS;                            
                                                                                 
-lookup mark {
```
khaledhosny, line 3: Is `mark` not a valid lookup name now?

tests/makeotfexe_data/input/spec/3-2.fea
```diff
@@ -1 +1 @@
-include ( 2gi-1.fea )
+include ( 2gi-1.fea ) ;
```
khaledhosny, line 1: Is the semicolon required now?

tests/makeotfexe_data/input/spec/8a-11.fea
```diff
@@ -0,0 +1,13 @@
+# When the replacement glyph of a single substitution is the same as
+# the target glyph of a multiple substitution, the former is encoded
+# as an alternate rather than a single substitution. That should happen
+# with A by A.sc and C by C.sc below.
```
khaledhosny, line 4: Is this a new behavior or an old one?

tests/makeotfexe_data/input/spec/3-2.fea
iterumllc: Yes, semicolons are required after include statements.

docs/OpenTypeFeatureFileSpecification.md
```diff
@@ -825,14 +826,37 @@ DEU
...
+    +  # Plus sign
+    :  # Colon
+    ?  # Question mark
+    ^  # Caret
+    '  # Back-quote
+    |  # Vertical bar
+    ~  # Tilde
+
+and must not start with a digit or hyphen. However, use of characters beyond
+those in production glyph names is not recommended.
```
khaledhosny, line 850: OpenType tags have [different requirements](https://docs.microsoft.com/en-us/typography/opentype/spec/otff#data-types) than what is stated here:
> A Tag value is a uint8 array. Each byte within the array must have a value in the range 0x20 to 0x7E. This corresponds to the range of values of Unicode Basic Latin characters in UTF-8 encoding, which is the same as the printable ASCII characters. As a result, a Tag value can be re-interpreted as a four-character sequence, which is conventionally how they are referred to. Formally, however, the value is a byte array. When re-interpreted as characters, the Tag value is case sensitive. It must have one to four non-space characters, padded with trailing spaces (byte value 0x20). A space character cannot be followed by a non-space character.

tests/makeotfexe_data/input/bug993/feat.fea
iterumllc: This is presumably negotiable. `mark` is a keyword and has been special-cased as a valid tag. Labels are arbitrary strings that aren't output into generated fonts, so there's no reason for keyword/label overlap. However, if there is enough history it can be added back in.

tests/makeotfexe_data/input/spec/8a-11.fea
iterumllc: This is a test of behavior in both the old feat.c and new FeatCtx.cpp (implemented in `aaltCreate()` in both cases).

docs/OpenTypeFeatureFileSpecification.md
iterumllc: Technically tags with that degree of generality weren't supported by the previous implementation -- it skipped the semicolon.

Supporting *unescaped* tags with that full range would either require going back to a mixed parser-lexer (an anti-goal of this port) or would greatly complicate the grammar. It was felt that the more limited set would be fine in practice.

An alternative, which would be straightforward to add, would be to also allow strings in the tag position for tags beyond those in the more limited character set (with escaped quotes if necessary).

tests/makeotfexe_data/input/spec/3-2.fea
iterumllc: (The addition of this test to the `spec` directory wasn't meant to have any strong implication. It's not clear that 8.a requires it. `makeotfexe` apparently does, though, so it needed a test.)

tests/makeotfexe_data/input/spec/8a-11.fea
iterumllc: (The addition of this test to the spec directory wasn't meant to have any strong implication. It's not clear that 8.a requires it. `makeotfexe` apparently does, though, so it needed a test.) (Edit: moved this to right thread.)

tests/makeotfexe_data/input/spec/3-2.fea
khaledhosny: It is a breaking change and needs a good justification (and documentation). It is a known fact that `makeotfexe` [allows `include` statements without terminating semicolon](https://github.com/adobe-type-tools/afdko/issues/114), and other implementations already [replicate this behavior](https://github.com/fonttools/fonttools/pull/642).

tests/makeotfexe_data/input/spec/8a-11.fea
khaledhosny: Good to know.

docs/OpenTypeFeatureFileSpecification.md
khaledhosny: At least the new implementation should support the same tags supported by the previous implementation, otherwise it would be a breaking change and would limit makeotf from accepting otherwise valid OpenType tags (it might not be an real issue now, but who knows when such tags would be required in the future).

tests/makeotfexe_data/input/bug993/feat.fea
khaledhosny: It is also a breaking change, and the fact that two tests already used `mark` as lookup names would be an indication that it is somehow a normal thing to do. I often name lookups with the feature name they will be used in, and it has always been accepted.

tests/makeotfexe_data/input/spec/3-2.fea
frankrolf: This has been part of a deprecation message for a while now …

iterumllc:
@khaledhosny Note that my comments were just attempting to clarify some of the thinking leading to this PR. I'll make similar notes on new issues as they are raised but otherwise let you and the reviewers work out these questions.

tests/makeotfexe_data/input/spec/3-2.fea
khaledhosny: `except` statement has been deprecated for more than a decade and it is still supported in this iteration.

iterumllc:
The already-high commit count almost doubled due to the Github workflow debugging so I squashed the PR for the sake of clarity. It doesn't look like Github managed to re-associate the existing review comments but the discussion has enough context to trace back to the relevant files.

docs/OpenTypeFeatureFileSpecification.md
josh-hadley: @khaledhosny regarding breaking changes: we recognize this, and for that reason have opted to bump the version number of the spec and have added appropriate documentation. An implementation that prefers the older scheme is free to peg their level of support to an older version of the spec (or use an older version of AFDKO/makeotfexe that implements the older spec).

tests/makeotfexe_data/input/bug993/feat.fea
josh-hadley: @khaledhosny see https://github.com/adobe-type-tools/afdko/pull/1367#discussion_r646953030

tests/makeotfexe_data/input/spec/3-2.fea
josh-hadley: @khaledhosny see https://github.com/adobe-type-tools/afdko/pull/1367#discussion_r646953030

docs/OpenTypeFeatureFileSpecification.md
khaledhosny: Fair enough.

josh-hadley: It does seem appropriate for us to note these changes prominently once we get this merged and into the release process.

tests/makeotfexe_data/input/bug993/feat.fea
josh-hadley: Good news! @iterumllc has kindly made an update (https://github.com/adobe-type-tools/afdko/pull/1367/commits/35f29a2b385f9696ffd072530758d535070179a2) that retains `mark` as a valid lookup name.

josh-hadley (approved):
Thanks so much @iterumllc for this great set of improvements!

PR #1370, opened 2021-06-04
tx uforead deleting outlines fix

### LINKED ISSUE #1366: [uforead] Some outlines are getting wiped out during UFO read
After commit https://github.com/adobe-type-tools/afdko/commit/a7f3917ea47ad8007f6b8b1096f6c8d6b5e1f9f2 I can't run a specific variable build process anymore because some outlines are randomly wiped out. It happens in at least one master and then the variable font can't be built because there are no more outlines to interpolate.

kaydeearts:
This should fix #1366.


josh-hadley (approved):
LGTM

cjchapman (approved):
👍

PR #1372, opened 2021-06-09
Copy token string when using FeatCtx::tokstr()

iterumllc:
## Description

I had thought that `std::string` objects retrieved from the parser tree are valid for the lifetime of the tree but there seems to be some other smart-pointer stuff going on. In most cases there is no difference (because strings are either used in the function where they're retrieved or passed "down") but FeatCtx::tokstr() needs to be changed to store the string.

I also changed the "reference token" for feature blocks to the start tag so that the error message makes sense. (That would have been a better choice to begin with anyway.)

## Checklist:


kaydeearts (approved):
👍  Thanks for this quick fix!

PR #1377, opened 2021-06-11
[build_wheels.yml] Try yum instead of apt-get for linux

kaydeearts:
Related to #1376 , #1375 , #1374 

Attempting to fix `sh: apt-get: command not found`, which may be because of docker linux images using `yum` as a package manager.


cjchapman (approved):
👍

PR #1380, opened 2021-06-14
[tx] Move FDArray from fontinfo.plist to lib.plist

kaydeearts:


c/shared/source/ufowrite/ufowrite.c
```diff
@@ -649,6 +550,159 @@ static void writeBlueValues(ufwCtx h, abfPrivateDict *privateDict){
...
+        glyphRec = &h->glyphs.array[i];
+        sprintf(buffer, "\t\t<string>%s</string>", glyphRec->glyphName);
+        writeLine(h, buffer);
+    }
+
+    writeLine(h, "\t</array>");
+
+    if (h->top->sup.flags & ABF_CID_FONT) {
+        if (h->top->cid.CIDFontName.ptr != NULL) {
+            writeLine(h, "\t<key>com.adobe.type.cid.CIDFontName</key>");
```
punchcutter, line 639: Can we drop `cid` from all the keys so just `com.adobe.type.CIDFontName`, etc? Originally I was thinking `cid` would make it clearer, but the rest of the info already does that.

punchcutter:
After looking at https://github.com/adobe-type-tools/afdko/issues/1396 I wonder did we up the version numbers for anything yet?

kaydeearts:
@punchcutter In my new commits I bumped the version numbers & fixed uforead for non-cid-keyed UFOs with glifs named `cidXXXXX`. I added a test for it too, does that test reflect the same issue you ran into?

punchcutter:
`tx` on this UFO fails with a segmentation fault because of the `<integer>` in `lib.plist` here: 
```
    <key>designspace.location</key>
    <array>
      <array>
        <string>weight</string>
        <integer>95</integer>
      </array>
    </array>
```
If I move this section to a different location it will work fine, but right after `FSType` or `Supplement` will fail because those are also expecting `<integer>`.  

Another issue is the warning `tx: (ufr) Warning: glyph order does not contain glyph name 'cid00000'` for every glyph in the font because the `public.glyphOrder` gets written at the end of the file and we are expecting to have read it first. 

[cidtest-ExtraLight.ufo.zip](https://github.com/adobe-type-tools/afdko/files/6931269/cidtest-ExtraLight.ufo.zip)

kaydeearts:
Changes in this PR will need to be redone due to the [replacing tx parser with libxml2 project. ](https://github.com/adobe-type-tools/afdko/projects/7)
Keeping this as reference for when I work on this on a new branch.

kaydeearts:
Closing due to #1576 completing this work.

PR #1383, opened 2021-06-15
[build_wheels.yml] Update to manylinux2014 (CentOS7), use BEFORE_ALL_LINUX

kaydeearts:


cjchapman (approved):
👍

anthrotype:
just curious, did something not work with manylinux1 that prompted the switch to the 2014 manylinyx image? the former wheel tag has wider support than the latter (older pips and linux versions)

kaydeearts:
@anthrotype Yes, we switched from manylinux1 to manylinux2014 after running into problems with the `libuuid-devel` library being installed but not found within the wheel. Since manylinux1 is based on CentOS5 which was deprecated in 2017, updating to the current manylinux2014 tag (based on CentOS7) fixed that problem.

anthrotype:
ok thanks for the clarification

PR #1384, opened 2021-06-16
Add more detailed docs for building cmake-based AFDKO from source

iterumllc:
## Description

The Antlr 4 Cpp runtime adds a guid/uuid to some elements of its parsers. It can satisfy that dependency with built-in parts of macOS and Windows but on other OSes `libuuid` and its headers must be installed (it uses `pkg_check_modules(UUID REQUIRED uuid)` in `CMakeLists.txt` on Linux).

This is effectively another "build tools" requirement and will already be satisfied on many systems, but some users will need to install an extra package or, possibly in rare cases, install libuuid from source. This PR adds some documentation to that effect. There's also some doc cleanup here and there.  

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,8 @@ Note for macOS users: we recommend that you do **not** use the system Python. Am
 
 Note for all users: if you are in a Python 3 only environment, then the command `pip` is sufficient.  If you are in a mixed Python 2 and Python 3 environment, or you are unsure of your environment, then the command `pip3` ensures that you are using the Python 3 version of `pip`. It is for this reason that we have used `pip3` in the instructions below.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -107,8 +109,15 @@ On Linux (Ubuntu 17.10 LTS or later), install these with:
     apt-get -y install python3.6
     apt-get -y install python-pip
     apt-get -y install python-dev
+    apt-get -y install uuid-dev
+
+On other POSIX-like operating systems, `libuuid` and its header files
+may be in a package named `libuuid-devel` or `util-linux-libs`. The
```
cjchapman, line 115: You also mentioned `uuid-dev` below, should it be mentioned here too?

cjchapman: Oh, nevermind, I see the `apt-get -y install uuid-dev` above now.

iterumllc: For whatever reason the current docs focus on ubuntu-family distros. I searched for the package names for the major families and these three are what I came up with (`util-linux-libs` is the post-May-2020ish name on Arch).

kaydeearts (dismissed):
👍  Looks good!

josh-hadley (approved):
👍

PR #1391, opened 2021-07-24
makeinstancesufo fixes for fontMath 0.8.1 update

kaydeearts:
This fixes test values for new, more accurate `italicAngle` values produced by the new fontMath 0.8.1 update.
This also fixes a bug in `makeinstancesufo` (not caught in tests until now) where `roundSelectedFontInfo(dFont.info)` was only called when the -r option was used (because it would have no-rounding _except_ for the values that need it) when they should always be rounded


josh-hadley (approved):
👍

PR #1394, opened 2021-07-26
[build_wheels.yml] install from requirements for specific dependency …

kaydeearts:
Same as #1393 but for build_wheels.yml


josh-hadley (approved):
LTGM

PR #1452, opened 2021-12-23
[tx] preserve intentional duplicate start points when PFA -> UFO

kaydeearts:
Duplicate points of type line should not be removed in `tx` when writing a UFO. These points are needed for variable fonts.


requirements.txt
```diff
@@ -2,7 +2,7 @@
 # we want to have control over the version and guarantee that the XML output
 # of our tools is stable
 # NOTE: hard-pinning (==) here gets relaxed to >= in setup.py
-lxml==4.7.1
```
josh-hadley, line 5: I think we want to keep the newer lxml now

josh-hadley (changes requested):

requirements.txt
kaydeearts: Oh I missed that during the rebase, just fixed. Thanks for catching!

josh-hadley (approved):
👍

PR #1459, opened 2022-01-05
[spec] Clarify nameid syntax and implementation details

### LINKED ISSUE #1457: how to edit name table id 5?
[According to the docs](http://adobe-type-tools.github.io/afdko/OpenTypeFeatureFileSpecification.html#9e-name-table), `name` table overrides can be specified within a feature file. I have found, however, that `makeotf` complains about attempts to override `name` ID 3, 4, or 5 (the font compiles but `makeotf` issues a warning that “name id 5 cannot be set from the feature file”).

1. Is the documentation wrong, or is this a bug in `makeotf`?

2. Is there another way of editing these `name` fields in a way that `makeotf` will tolerate?

PS. `makeotf` itself automatically generates an entry for `name` ID 5 that looks like "Version 2.548;hotconv 1.0.116;makeotfexe 2.5.65601".

josh-hadley:
## Description

This is an update to the documentation (only) to clarify some details about `nameid` syntax and to distinguish some `makeotf`/`makeotfexe` implementation details from the core specification.

- re-work table explaining `id` and `string attributes`
- highlight the behavior of MakeOTF/makeotfexe implementations regarding overriding of certain default/calculated nameIDs via feature file

Closes #1457 

## Checklist:

- ~[ ] I have added **test code and data** to prove that my code functions correctly~
- ~[ ] I have verified that new and existing tests pass locally with my changes~
- ~[ ] I have commented my code, particularly in hard-to-understand areas~

### DIFF
--- a/docs/OpenTypeFeatureFileSpecification.md
+++ b/docs/OpenTypeFeatureFileSpecification.md
@@ -10,8 +10,8 @@ Copyright 2015-2021 Adobe. All Rights Reserved. This software is licensed as
 OpenSource, under the Apache License, Version 2.0. This license is available at:
 http://opensource.org/licenses/Apache-2.0.
 
-Document version 1.26
-Last updated 7 June 2021
+Document version 1.27
+Last updated 5 January 2022
 
 **Caution: Portions of the syntax unimplemented by Adobe are subject to change.**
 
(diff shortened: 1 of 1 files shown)


mbutterick:
* The term “implementation” is used in an ambiguous manner throughout the spec. It can mean either “any implementation out in the world” or “Adobe’s particular implementation in `makeotf`”. This is confusing.

* It seems like it would be more fruitful to improve `makeotf` to better implement the spec rather than patch the spec documentation to explain its shortcomings)

* If that’s not possible, then I think all the `makeotf` material in the OTFF spec should be more clearly and consistently labeled. Among other things, rather than “Implementation Notes” it should say “`makeotf` notes”. Instead of “Portions of the syntax unimplemented by Adobe” it should read “portions of the syntax unimplemented by Adobe in `makeotf`”. And so on.

mbutterick:
* The `overrideMenuNames` flag is a misnomer, since it can be used to override the version string, which is not a menu name.

josh-hadley:
@mbutterick sorry this PR does not meet your expectations. From your comments, it seems that a larger overhaul of the spec and tool(s) is needed beyond the concerns raised in #1457. Unfortunately that's out-of-scope for the time being, but perhaps some future efforts can address it. I'll go ahead and close this without merging and hope to return to a more extensive re-work of things eventually.

PR #1469, opened 2022-02-03
[requirements.txt] Update psautohint requirements

kaydeearts:


josh-hadley (changes requested):
Looks like fontTools just recently went to [4.29.1](https://github.com/fonttools/fonttools/releases/tag/4.29.1), let's pick that up also

josh-hadley (approved):
Looks good!

PR #1478, opened 2022-03-08
[tx] Don't parse non-FDArray array of dictionaries in fontinfo.plist

kaydeearts:
Masataka reported that he found an error when using `mergefonts`. The error is actually in `tx`.
Because the UFO was exported from Glyphs3, there was some additional data within fontinfo.plist that was accidentally parsed as FDArray. 

For now, I've added a `parsingFDArray` bool value that is set to `true` only if the `postscriptFDArray` key was seen when parsing an array of font dictionaries. This is a **temporary solution**, since the switch to libxml2 parsing will have a more elegant way of dealing with this. 

We need to decide if it's necessary to start storing this kind of additional data from Glyphs 3 and other apps in the UFO when running it through our tools.


josh-hadley (approved):
Looks good (and I'm looking forward to a more permanent solution with `libxml` 😄 )

PR #1483, opened 2022-03-15
git protocol security fix for utfcpp/antlr4

kaydeearts:
This is a temporary fix for installing utfcpp within ANTLR4 build. GitHub's protocol security improvement was made permanent today: https://github.blog/2021-09-01-improving-git-protocol-security-github/

The fix right now is running `git config --global url.https://github.com/.insteadOf git://github.com/` in the wheels before building the FDK. We had these commands in the wheels before, but were removed a few months ago.
The permanent fix for this will be upgrading to the latest ANTLR4.


josh-hadley (approved):
🦭

PR #1484, opened 2022-03-15
[tx] ignore subr recursion limit

### LINKED ISSUE #1482: [tx] conditionally ignore subroutine recursion limit
A font whose CFF table exceeds the subroutine recursion limit cannot be dumped by `tx` using `tx -cff -S fontfile.otf > table.CFF` (result is a zero-length file and an error message from `tx`: "subr recursion level limit exceeded"). Thus it is not possible to repair such a font using desubroutinize + re-subroutinize procedures documented in the [How-to](https://github.com/adobe-type-tools/afdko/wiki/How-To#how-to-de-subroutinize-a-cff-or-cff2-font) section of the Wiki.

If `tx` could ignore the limit (perhaps with an additional flag for `-cff` mode, or infer when using `-S`), it would allow repairing such fonts by de-subroutinizing (`-S`), then re-subroutinizing (`+S`) the CFF table.

josh-hadley:
## Description

This PR updates t2cstr which handles various `tx -cff` duties. Previously, when dumping a font whose CFF table exceeded subroutine (or gsubr) recursion limits, `tx` would emit an error message and generate a zero-length file. This update still prints a message, but clarifies that it is ignoring the condition (as is the case for several other out-of-bounds font values) and continues the dump of the out-of-bounds data. It also updates the formatting of the message slightly.

Fixes #1482 

## Checklist:

- [ ] ~I have added **test code and data** to prove that my code functions correctly~ I haven't done this because the only test data I have available is a private font which cannot be added to this open source repo.


kaydeearts (dismissed):
👍

miguelsousa:
Should this line get removed?
https://github.com/adobe-type-tools/afdko/blob/5783516994a7db1f4bb41b3d441682aacbd61478/c/shared/include/t2cerr.h#L23

josh-hadley:
> Should this line get removed?
 
Yes, nothing references that now. Done in https://github.com/adobe-type-tools/afdko/pull/1484/commits/82c4128ab691dc3eb9b5e00fb2ef7ba853e3d49d . Thanks for catching that 😄

PR #1491, opened 2022-03-18
ANTLR v4.9.3 Upgrade

kaydeearts:


josh-hadley (approved):

.github/workflows/build_wheels.yml
```diff
@@ -43,7 +43,6 @@ jobs:
         output-dir: dist
       env:
         CIBW_BUILD: "cp39-*"
-        CIBW_BEFORE_ALL_MACOS: "git config --global url.https://github.com/.insteadOf git://github.com/"
```
iterumllc, line 46: Should this be removed? (You know better than I -- just checking.)

iterumllc (approved):
Other than the question LGTM.

.github/workflows/build_wheels.yml
kaydeearts: Yes this should be fine to remove — I added this as a temporary fix to the utfcpp download problem with the previous ANTLR version. Thanks for checking!We'll really find out if there's problems with this yml file once I merge this and push a new tag

PR #1494, opened 2022-03-28
[tx] memory leak fixes

kaydeearts:
This PR fixes memory leaks found on OpenSUSE86 with these flags added to root-level CMakeLists.txt:

`add_compile_options(-fsanitize=address)`
`add_link_options(-fsanitize=address)`


c/shared/source/uforead/uforead.c
```diff
@@ -401,12 +401,61 @@ void ufoFree(ufoCtx h) {
...
+    for (int i = 0; i < h->data.opList.cnt; i++){
+        OpRec rec = h->data.opList.array[i];
+        if (rec.pointName != NULL)
+            memFree(h, rec.pointName);
+    }
     dnaFREE(h->data.opList);
     freeStrings(h);
     dnaFree(h->dna);
 
+    if (h->top.cid.CIDFontName.ptr != NULL)
```
skef, line 433: I haven't gone too far in depth with the analysis but I'm worried about these lines and the shared responsibility over the `top` structure. 

First and most basically it seems like if this cleanup is going to happen at all it should be in a function in `shared/source/absfont/absfont.c`, probably called `abfFreeTopDict()` to match `abfInitTopDict()`. 

But that's a detail. The more pressing problem is that `top` is typically passed into `ufoBegFont()` as a double-pointer, which is then initialized and also stored in the `ufoCtx`. This is not ideal to begin with -- it would be preferable if `ufoBegFont()` could set the UFO `h->top` back to NULL before returning -- but if the ctx will still be used and needs it that may be impractical.

Let's assume that the double-pointer `top` is only valid a) while the ufoCtx is allocated and b) between calls to `ufoBegFont()` (It seems like one can call `ufoEndFont()` and not loose anything, as that's just for closing files and such?) 

In that case the ufoCtx top structure may be re-used (right? there's only one in the `txCtx` ...), and that's a problem given that `abfInitFontDict()` does't check for and free already-allocated values. A normal thing here would be to have `abfReInitTopDict()` which calls `abfFreeTopDict()` first and then `abfInitTopDict()`, and just use the `ReInit` version when you may be in a reuse situation. But the code could also just call `abfFreeTopDict()` first. 

Doing this safely is going to require a bit of analysis of the specific lifetime of the `top` structure.

All of these comments also apply to `fdict`.

c/shared/source/uforead/uforead.c
```diff
@@ -954,67 +1003,91 @@ static void setFontDictKey(ufoCtx h, char* keyValue) {
     /* If we do not use a keyValue, we need to dispose of it
      */
     if (!strcmp(keyName, "copyright")) {
-        top->Copyright.ptr = keyValue;
+        top->Copyright.ptr = memNew(h, strlen(keyValue) + 2);
```
skef, line 1006: Why do all of these strings need an extra byte?

PR #1497, opened 2022-04-06
[uforead.c] Fix Memory Crashes

kaydeearts:
Fixing memory crashes found in OpenSUSE x86 VM and Mac (with address sanitizer). 

Each FDict in FDArray needed to be initialized after memory was allocated


skef (approved):
Looks good to me

PR #1501, opened 2022-04-21
chore(ci): add informational Codecov status checks

thomasrockhu-codecov:
## Description
Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. Let me know if this makes sense or if we can do something that would be helpful.

## Checklist:

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,8 +2,12 @@ comment: false
 
 coverage:
   status:
-    project: off
-    patch: off
+    project:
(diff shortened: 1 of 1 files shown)


kaydeearts (approved):
Thank you for your contribution! We'll try this out 👍

Issue #1467, opened 2022-02-01
`tx -dump` fails if fontinfo.plist contains `openTypeNameRecords` key


frankrolf:
Found out while attempting `makeotf` on an arbitrary UFO – the error is
```
Command '['tx', '-dump', '-0', 'NameRecordsTest.ufo']' returned non-zero exit status 6.
```

Trying 
`tx -dump -0 NameRecordsTest.ufo` directly, the error is
```
tx: --- NameRecordsTest.ufo
tx: (ufr) Warning: discarding token '</plist', in fontinfo.plist file. Context: 'g>
	</dict>
</plist>'.
tx: (ufr) Encountered end of buffer before end of glyph.g>
	</dict>
</plist>.
```

It seems `tx` simply does not expect this kind of data structure.

frankrolf:
A UFO exhibiting this behavior is attached – barely anything in it, except the OT name records in the `fontinfo.plist`.
[NameRecordsTest.ufo.zip](https://github.com/adobe-type-tools/afdko/files/7982197/NameRecordsTest.ufo.zip)

josh-hadley:
> tx simply does not expect this kind of data structure
This seems to be the case. The ufo read code is not expecting an `<array>` for the value of a `<key>` that it doesn't know about. It seems to assume only one-level values like `<string>` and so on (If I rip out the whole `<array>` block and substitute with `<string>Foo</string>` following that key, it doesn't fail).

punchcutter:
This is one of many variations on this. I ran into a ton more, but the solution is to use libxml2 or something instead of this hard coded XML parsing.

kaydeearts:
This is fixed by #1515 and I've added a [test case to tx_test.py.](https://github.com/adobe-type-tools/afdko/blob/c05fa225bf9a3060b2ea79eabaaf70352b1890b1/tests/tx_test.py#L1255-L1264)

frankrolf:
Excellent news, thank you @kaydeearts! :-) I assume #1396 might also be fixed by this?

PR #1510, opened 2022-05-12
[tx] Replace UFO fontinfo.plist parsing with libxml2

kaydeearts:
## Description

This is part of a larger project to replace current xml parsing code in `tx` with the libxml2 library. 

This PR replaces fontinfo.plist parsing code with libxml2, as well as sets up the AFDKO for cross-platform compatibility with libxml2 (see CMake changes)

This should also resolve issue #1467 , for which a test has been added in tx_test.py

## Checklist:


.github/workflows/build_wheels.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       env:
         CIBW_BUILD: "cp39-*"
         CIBW_ARCHS_MACOS: universal2
-        CIBW_ENVIRONMENT_MACOS: "CFLAGS='-arch arm64 -arch x86_64' CXXFLAGS='-arch arm64 -arch x86_64' LDFLAGS='-arch arm64 -arch x86_64'"
+        CIBW_ENVIRONMENT_MACOS: "CFLAGS='-arch arm64 -arch x86_64 -I/usr/include/libxml2 ./configure' CXXFLAGS='-arch arm64 -arch x86_64' LDFLAGS='-arch arm64 -arch x86_64'"
```
skef, line 47: Does this "./configure" still have a role?

CMakeLists.txt
```diff
@@ -42,6 +42,15 @@ set(ANTLR4_WITH_STATIC_CRT OFF)
 set(ANTLR4_TAG tags/4.9.3)
 include(ExternalAntlr4Cpp)
 
+FIND_PACKAGE(LibXml2)
+IF (WIN32)
```
skef, line 46: Doing this on a WIN32 basis will work fine for our internal build purposes, but in the field a given box may or may not have libxml2 on it already, so a search (probably using `find_package`) will make more sense. 

I thought I had seen some messages that we were already doing that. Was that in a different draft/branch?

c/makeotf/lib/cffread/CMakeLists.txt
```diff
@@ -1,4 +1,13 @@
 add_library(makeotf_cffread STATIC cffread.c)
 
-target_compile_definitions(makeotf_cffread PRIVATE $<$<CONFIG:Debug>:CFF_DEBUG=1> CFF_T13_SUPPORT=0)
+target_include_directories(makeotf_cffread PRIVATE AFTER $<$<COMPILE_LANGUAGE:CXX>:${ANTLR4_INCLUDE_DIRS}>)
+
+if(WIN32)
+    target_link_libraries(makeotf_cffread PUBLIC wsock32 ws2_32 libxml2)
```
skef, line 6: Given that they're repeated so often and could change with different compilers or C libraries it would probably better to use a variable to specify these three somewhere and refer to that.

c/shared/source/tx_shared/tx_shared.c
```diff
@@ -405,6 +405,31 @@ static size_t stm_read(ctlStreamCallbacks *cb, void *stream, char **ptr) {
     return 0; /* Suppress compiler warning */
 }
 
+static size_t stm_xml_read(ctlStreamCallbacks *cb, void *stream, char **ptr, xmlDocPtr *doc){
+    int res, size = BUFSIZ;
+    char chars[BUFSIZ];
```
skef, line 410: Confused about `chars` and `s->buf`. I don't see where we use the former and therefore am worried about the relation between `size` (which we set here`) and `s->buf` (which presumably comes in with `stream`?).

c/shared/source/tx_shared/tx_shared.c
```diff
@@ -405,6 +405,31 @@ static size_t stm_read(ctlStreamCallbacks *cb, void *stream, char **ptr) {
...
+    xmlParserCtxtPtr ctxt;
+
+    Stream *s = stream;
+    txCtx h = cb->direct_ctx;
+    if (h->seg.refill != NULL)
+        return h->seg.refill(h, ptr);
+
+    *ptr = s->buf;
```
skef, line 418: Not sure `ptr` will point to something sensible after we go though this loop (given that `buf` gets re-used), and if that's right its concerning that we're passing it back from here.

c/shared/source/uforead/uforead.c
```diff
@@ -953,174 +956,170 @@ static void setFontDictKey(ufoCtx h, char* keyValue) {
...
+        parsingFDArray = false;
+    } else if (!strcmp(keyName, "PrivateDict")) {
+        parsingFDArray = false;
+        parseKeyValue(h, cur);
+        parsingFDArray = true;
+    } else {
+        char* keyValue = (char*) parseKeyValue(h, cur);
+        if (keyValue != NULL && !strcmp(keyValue, ""))
+            message(h, "Warning: Encountered empty <%s> for fontinfo key %s. Skipping", cur->name, keyName);
+        if (!strcmp(keyName, "copyright")) {
```
skef, line 978: Looks like from a few samples we've only got the additions above and some indentation changes below so I'm reviewing on that basis. Let me know if there are more substantive changes among the indentations that I should look at.

c/shared/source/uforead/uforead.c
```diff
@@ -1819,9 +1818,153 @@ static void reallocFDArray(ufoCtx h){
...
+        fprintf(stderr, "document of the wrong type, root node != %s", filetype);
+        xmlFreeDoc(doc);
+//        return(NULL);
+    }
+
+    cur = (cur)->xmlChildrenNode;
+    while (cur && xmlIsBlankNode(cur)) {
+        cur = (cur) -> next;
+    }
+    if ( cur == 0 ) {
```
skef, line 1851: Better to check pointers against NULL. Also, given that the return is commented out this seems unsafe -- we're about to use `cur->name` in a test.

c/shared/source/uforead/uforead.c
```diff
@@ -1819,9 +1818,153 @@ static void reallocFDArray(ufoCtx h){
...
+        xmlFreeDoc(doc);
+//        return ( NULL );
+    }
+    if ((!xmlStrEqual((cur)->name, (const xmlChar *) "dict"))) {
+        fprintf(stderr, "Error reading outermost <dict> in %s.\n", "file");
+        xmlFreeDoc(doc);
+//        return(NULL);
```
skef, line 1858: Similarly unsafe-seeming

c/shared/source/uforead/uforead.c
```diff
@@ -1819,9 +1818,153 @@ static void reallocFDArray(ufoCtx h){
...
+        cur = cur->next;
+    }
+}
+
+static void parseKeyName(unsigned char* *keyID, xmlNodePtr cur){
+    cur = cur->xmlChildrenNode;
+    while (cur != NULL) {
+        if (!xmlStrEqual(cur->name, (const xmlChar *) "text"))
+            continue;
+        *keyID = xmlNodeGetContent(cur);
```
skef, line 1884: If we want to use `keyID` as a char, maybe just cast it here?

Also, should we just `return` here? Two "text" nodes is unlikely but if it happens we'll probably leak the first assignment.

c/shared/source/uforead/uforead.c
```diff
@@ -1819,9 +1818,153 @@ static void reallocFDArray(ufoCtx h){
...
+        keyID = NULL;
+        if ((xmlStrEqual(cur->name, (const xmlChar *) "key"))) {
+            parseKeyName(&keyID, cur);
+            if (keyID == NULL){
+                message(h, "Warning: Encountered empty <key></key>.");
+                cur = cur->next;
+                continue;
+            }
+            keyName = (char*) keyID;
+            cur = cur->next;
```
skef, line 1872: This could well be right but could you briefly explain why we're advancing twice when we read a key with a parseable name and once when it's not a key or we can't parse the name?

c/shared/source/uforead/uforead.c
```diff
@@ -1819,9 +1818,153 @@ static void reallocFDArray(ufoCtx h){
...
+static char* parseKeyError(ufoCtx h, xmlNodePtr cur){
+    if (xmlStrEqual(cur->name, (const xmlChar *) "key"))
+        fatal(h, ufoErrParse, "Encountered empty <key>");  // needs fix
+    else if (xmlStrEqual(cur->name, (const xmlChar *) "array"))
+        fatal(h, ufoErrParse, "Encountered empty <array>");
+    else
+        return NULL;
+}
+
+static void *parseKeyValue(ufoCtx h, xmlNodePtr cur){
```
skef, line 1947: Even though it can be convenient it's best to avoid `void *` unless and until absolutely necessary (that is, don't do it if you don't have to, and if you do need a void pointer don't cast something to it until as late as possible). 

Seems like this function is mostly intended to return a string. Are there other cases?

c/shared/source/uforead/uforead.c
```diff
@@ -1819,9 +1818,153 @@ static void reallocFDArray(ufoCtx h){
...
+    if (isSimpleKey(cur)) {  /* if string, integer, or real */
+        void *ptr = xmlNodeGetContent(cur);
+        return ptr;
+    } else if (xmlStrEqual(cur->name, (const xmlChar *) "dict")) {
+        parseXMLDict(h, cur);
```
skef, line 1952: We never define what we're returning here, as far as I can see. We just slip through the rest of the guards to the end of the function.

c/shared/source/uforead/uforead.c
```diff
@@ -1819,9 +1818,153 @@ static void reallocFDArray(ufoCtx h){
...
+        *keyID = xmlNodeGetContent(cur);
+        cur = cur->next;
+    }
+}
+
+static void parseXMLArray(ufoCtx h, xmlNodePtr cur){
+    dnaSET_CNT(h->valueArray, 0);
+    cur = cur->xmlChildrenNode;
+    while (cur != NULL) {
+        void* valueString = parseKeyValue(h, cur);
```
skef, line 1893: See note about `void *` below.

c/shared/source/uforead/uforead.c
```diff
@@ -1819,9 +1818,153 @@ static void reallocFDArray(ufoCtx h){
...
+        if (h->top.FDArray.cnt > FDArrayInitSize){ /* Memory needs reallocation*/
+            reallocFDArray(h);
+        }
+        abfFontDict* fd = h->top.FDArray.array + currentiFD;
+        abfInitFontDict(fd);
+    }
+
+    while (cur != NULL) {
+        parseKeyName(&keyID, cur);
+//        cur = cur->next;
```
skef, line 1917: Maybe just delete these comments now.

CMakeLists.txt
kaydeearts: Yes, that was a different branch. I think I put this back at the time to minimize complexity, but I can put it back now. Thanks!

c/makeotf/lib/cffread/CMakeLists.txt
kaydeearts: You're referring to `wsock32 ws2_32 libxml2`, right? 
I'll make a variable for these three, thanks 👍

skef: Yes, those dependencies

c/shared/source/tx_shared/tx_shared.c
kaydeearts: I looked into this a bit, but not sure if I understood. Is it the `return h->seg.refill(h, ptr); `line that's concerning, or the `*ptr = s->buf;`?

c/shared/source/uforead/uforead.c
kaydeearts: There are some general indent changes, but the biggest change for this function is at the top where now it first checks if it is at the beginning of parsing an FDArray. It checks for the postscriptFDArray key, and then does some initialization/etc for FDArray, before checking for any other values. I had to do it this way to make the FDArray parsing process recursive (I should add more documentation about this, so I'll do that).

kaydeearts: Another big change is the way keyName and keyValue is handled. Previously, keyValue was passed into the function, and keyName was held in h->parseKeyName.

c/shared/source/tx_shared/tx_shared.c
skef: The latter. This could easily be something I just don't understand. What is `ptr` supposed to represent as a return parameter?

c/shared/source/uforead/uforead.c
kaydeearts: When we read a nonekey or an 'empty' key (one that doesn't have a value defined for it), that one cur = cur -> next advances the document ptr to the next item (which will be the empty key value, or the next key if that key value was never defined). This value is then advanced over because it isn't a key, thus advancing again. So technically, it is also advanced twice.

When reading a key that are parseable: I set up the advancing so that I would call the parse functions, but still have the ptr at the same spot in case I wanted to do other things with it. To me, it seems like the caller should have ptr management responsibility. Do you think that makes sense?

c/shared/source/uforead/uforead.c
kaydeearts: Ah I see, that's good to know. You're right, it is fine to return either an unsigned char* or keep it as an XmlChar* (which is an unsigned char*). I would then have to cast this return value from unsigned to signed char* when assigning the dictionary values in setFontDictKey. DO you think there will be risks with that kind of conversion?

kaydeearts:
Self reminder: update versions

c/shared/source/uforead/uforead.c
kaydeearts: @skef says "I understand the basic idea here now. You'll check the potential case of loosing the second key of two keys in a row."

kaydeearts: @skef says "I understand the basic idea here now. You'll check the potential case of loosing the second key of two keys in a row."

skef (commented):
I know you may be in the middle of things but I thought I would have another quick look. This seems like it's getting in quite good shape!

c/shared/source/uforead/uforead.c
kaydeearts: I've refactored this part of the code (`parseXMLKeyName` and its callers) to clarify how `cur` is being advanced. You'll now see that the `keyName` validation happens within `parseXMLKeyName`.

I've also added tests to `tx_test.py::test_ufo_fontinfo_parsing` for cases like this, such as two keys back to back (with a value missing for the first one) to make sure we don't skip over the next key.

Based on those points, I think this can be resolved unless you have additional thoughts

c/shared/source/uforead/uforead.c
kaydeearts: All new uses of void* should now have been replaced by char*. 

I believe this can be resolved unless you have additional thoughts

PR #1518, opened 2022-06-15
[build_wheels.yml] Remove env installation of libxml2 for Windows

kaydeearts:
Attempting to fix pre-release fail where Windows cannot install libxml2 with these commands in the env (it needs to run ExternalProject_Add in cmake!). I think removing these commands should help, since the other wheels files (testpythonpackage.yml) build that way, but we'll have to test with another prerelease push


josh-hadley (approved):
Looks like that did the trick 👍

PR #1515, opened 2022-06-14
[tx] Replace UFO fontinfo.plist parsing with libxml2

kaydeearts:
This is the cleaned up version of #1510 
This is part of a larger project to replace current xml parsing code in `tx` with the libxml2 library. 

This PR replaces fontinfo.plist parsing code with libxml2, as well as sets up the AFDKO for cross-platform compatibility with libxml2 (see CMake changes)

This should also resolve issue #1467 , for which a test has been added in tx_test.py

## Checklist:

- [x] I have followed the [Contribution Guidelines](https://github.com/adobe-type-tools/afdko/blob/develop/CONTRIBUTING.md)
- [x] I have added **test code and data** to prove that my code functions correctly
- [x] I have verified that new and existing tests pass locally with my changes
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation

Issue #1396, opened 2021-07-31
[tx] trips over UFO with global guidelines


frankrolf:
Two UFOs attached – `without_guides.ufo` and `with_guides.ufo`.
The only difference is that the `with_guides.ufo` has a global guideline (which for some reason is stored in the `fontinfo.plist`, not the `lib.plist` – as I’d expect)

```
tx -t1 without_guides.ufo
```
→ 
```
normal T1 data is spit out
```

----

```
tx -t1 with_guides.ufo
```
→ 
```
tx: --- with_guides.ufo
tx: (ufr) Warning: Encountered empty <string> for fontinfo key name. Skipping
tx: (ufr) Warning: discarding token '</plist', in fontinfo.plist file. Context: 'r>
	</dict>
</plist>'.
tx: (ufr) Encountered end of buffer before end of glyph.r>
	</dict>
</plist>.
```

tx -v:
```
Versions:
    tx        1.2.4
    absfont   1.0.54
    ctutil    2.0.3
    dynarr    2.0.4
    sfntread  1.0.7
    t2cstr    1.0.22
    cffread   2.1.3
    cffwrite  1.0.56
    sfntwrite 1.0.6
    pstoken   2.0.11
    t1cstr    1.0.21
    t1read    1.0.45
    ttread    1.0.22
    cfembed   2.0.25
    pdfwrite  1.0.6
    svread    1.0.8
    t1write   1.0.35
    svwrite   1.1.11
    uforead   1.1.2
    ufowrite  1.0.9
    varread   1.0.8
```

[guide ufos.zip](https://github.com/adobe-type-tools/afdko/files/6910887/guide.ufos.zip)

punchcutter:
Works for me with the "same versions", but using this PR: https://github.com/adobe-type-tools/afdko/pull/1380
Better get those versions updated in the PR and then this will work.

frankrolf:
This does not seem to be solved (tangentially related to #1467):
`makeotf -f one_global_guide.ufo` → works
`makeotf -f two_global_guides.ufo` → Command `tx -t1 two_global_guides.ufo` returned non-zero exit status 5.
`tx -t1 two_global_guides.ufo ` → 
```
tx: --- two_global_guides.ufo
tx: (ufr) Warning: No PS name specified in source UFO font.
tx: (ufr) Warning: No version specified in source UFO font.
tx: (t1w) bad font dictionary. Missing the PostScript FontName.
```
(error message is misleading)

[...]

PR #1519, opened 2022-06-15
[tx_test] add test for bug1396

kaydeearts:
Adding test for #1396, which was fixed by #1515

Also rename a related test to specify it is to test fontinfo.plist parsing


josh-hadley (approved):
👍

Issue #1520, opened 2022-06-15
[tx] Replace UFO lib.plist/glyphOrder parsing with libxml2


kaydeearts:
Part of #1468 , #475. Need to replace old parser code for lib.plist

Issue #1305, opened 2021-02-15
makeotf: errors out if UFO lib.plist is not present


dy:
Trying to compile UFO folder as:
```
PS C:\projects\wavefont> makeotf -f wavefont.ufo -o font/wavefont.otf
Command '['tx', '-dump', '-0', 'C:\\projects\\wavefont\\wavefont.ufo']' returned non-zero exit status 6.
```
Is there some specifics to compiling UFOs?

frankrolf:
Nothing wrong here – there seems to be a problem with the UFO source data.
I don’t know what would return exit status 6, but what does `tx -dump -0 font.ufo` do?

dy:
Ok, I had comments in plist file as
```plist
<!-- For windows!-->
    <key>openTypeOS2WinAscent</key>
    <integer>1500</integer>
```
Which resulted in 
```
tx.exe: (ufr) Encountered end of buffer before end of glyph.openTypeOS2WinDescent</key>
    <integer>0</integer>

    <key>openTypeNameVersion</key>
    <string>Version 1.000</string>
  </.

```

So that was probably XML parser issue.

dy:
In fact just encountered less clear error.
Having a font created by fontforge and exported as UFO.
```
 makeotf -f ./test.ufo/ -o font/test.otf
Command '['tx', '-dump', '-0', 'C:\\projects\\wavefont\\test.ufo']' returned non-zero exit status 3.

tx -dump -0 ./test.ufo/
Failed to read lib.plist
tx.exe: --- ./test.ufo/
tx.exe: (ufr) source stream error
```

frankrolf:
This error helps me more.
FontForge does not write lib.plist, I don’t know why.

https://github.com/adobe-type-tools/afdko/issues/1130#issuecomment-623180851

dy:
`lib.plist` is optional at least in [UFO3](https://unifiedfontobject.org/versions/ufo3/lib.plist/):

> This file is a place to store authoring tool specific, user specific or otherwise arbitrary data for the font. It is optional. If it is not defined in the UFO, there is no lib data.

So that seems to be makeotf issue.

frankrolf:
see #1306

Issue #1130, opened 2020-05-03
Convert ufo that lack a lib.plist?


probonopd:
I would like to use afdko to convert a FontForge-generated `ufo` to `otf`. However, FontForge does not add a `lib.plist` to the `.ufo` directory. Is there a way to still convert it?

References:
* https://github.com/fontforge/fontforge/issues/2161
* https://github.com/probonopd/font-leutkirch/issues/2

frankrolf:
A workaround could be building the lib.plist (and the `public.glyphOrder` array within) yourself, based on the alphabetical order given in the normal contents.plist file. Here is a draft script:

```
'''
Create a lib.plist file in a given UFO, based on the information
in glyphs/contents.plist (alphabetical order).
'''

import os
import argparse
import plistlib

parser = argparse.ArgumentParser(
    description=__doc__)

parser.add_argument(
    'ufos',
    action='store',
    nargs='+',
    metavar='UFO',
    help='input UFO(s)')

args = parser.parse_args()

for ufo_path in args.ufos:
    contents_path = os.path.join(ufo_path, 'glyphs', 'contents.plist')
    lib_path = os.path.join(ufo_path, 'lib.plist')
    with open(contents_path, 'rb') as pl_file:
        contents_dict = plistlib.load(pl_file)

    glyph_names = sorted(contents_dict.keys())
    lib_dict = {}
    lib_dict['public.glyphOrder'] = glyph_names
    with open(lib_path, 'wb') as pl_file:
        plistlib.dump(lib_dict, pl_file)
```

probonopd:
This works flawlessly. Thank you so much @frankrolf.

frankrolf:
Great to hear! :-)

probonopd:
Do you think it may be integrated into afdko? Or should tools like FontForge do this?

frankrolf:
I think glyph order is quite important in a UFO file. I don’t see why an editing tool like FontForge would not be interested in writing the `public.glyphOrder` key.

That said, `tx` (which is the tool invoked while reading the UFO) should probably not fail if the lib.plist is not present.

josh-hadley:
Per [a related discussion](https://github.com/adobe-type-tools/afdko/issues/1255), I'm closing this because our policy is not to modify user UFOs.

ctrlcctrlv:
[...]

PR #1523, opened 2022-06-23
[tx] Replace UFO lib.plist parsing with libxml2

### LINKED ISSUE #1306: [makeotf] fail better when calling tx on UFO with missing lib.plist
We now have two issues (#1130 and #1305) which point at `tx` failing if the [optional](https://unifiedfontobject.org/versions/ufo3/lib.plist/) `lib.plist` file is missing in the UFO package. 

FontForge seems to be the only editor to write UFOs without `lib.plist`. Fonts with missing `lib.plist` file will also miss the `public.glyphOrder` key – which IMO is fairly essential.
I provided a Python workaround in #1130, but it would be nice for `tx`/`makeotf` to do that automatically, or (at least) provide a more helpful error message.

kaydeearts:
## Description

This PR replaces lib.plist parsing code with libxml2 #1520 .[ (part of the tx parser replacement with libxml2 project)](https://github.com/adobe-type-tools/afdko/projects/7), and is related to #1515 

This should also fix #1306 , #1130 , #1305 , which discuss that `tx`/`makeotf` should not fail when missing lib.plist. 
A missing lib.plist should now emit a warning to the user instead of failing. These warnings have been improved and tests have been added to both `tx_test.py` and `makeotf_test.py`. 
The only exception to this is for cidkeyed UFOs that have cid identifiers defined within the glyphs files, for which there was a failure message already existing. This test case has also been added.

## Checklist:


skef (dismissed):
Looks good to me!

josh-hadley (dismissed):
👍

PR #1526, opened 2022-06-28
[workflows] fix ubuntu wheel libxml2

kaydeearts:
## Description

Trying to fix wheel issue where libxml2 installation does not have `rpath` by using patch script to fix rpath

## Checklist:

### DIFF
--- a/.github/scripts/patchwheel.py
+++ b/.github/scripts/patchwheel.py
@@ -1,0 +1,62 @@
+"""
+This is a modified script for patching rpath,
+originally from https://github.com/arbor-sim/arbor
+"""
+
+import shutil
+import subprocess
+import argparse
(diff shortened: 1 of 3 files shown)


.github/scripts/patchwheel.py
```diff
@@ -0,0 +1,62 @@
+"""
```
skef, line 1: We can add the content of https://github.com/arbor-sim/arbor/blob/master/LICENSE as a comment in this file and throw a one-liner into afdko's LICENSE noting that. Or we can add the content to afdko's LICENSE file like I did with the Antlr CMake stuff.

skef (approved):
Looks good subject to a bit of LICENSE attention

PR #1527, opened 2022-06-29
[LibXML2] Link LibXML2 Statically for Linux

kaydeearts:
## Description

Due to a bug in the linux python wheels involving dynamically linking libxml2, discovered in #1525 , we'll be installing libxml2 externally through `ExternalLibXML2.cmake` and linking statically, just like we do in most Windows cases.
We will revisit this in the future and attempt to fix dynamic linking for linux.

## Checklist:

- [ ] I have followed the [Contribution Guidelines](https://github.com/adobe-type-tools/afdko/blob/develop/CONTRIBUTING.md)


josh-hadley (approved):

skef (approved):

josh-hadley (approved):

sternenseemann:
Can this please be made configurable? When packaging afdko for nixpkgs we can provide libxml2 via the package manager and be confident that it'll be available at runtime as well. What we don't want is to download and link a random version of libxml2 in when building afdko.

skef:
The Linux restriction in this release is likely temporary -- we needed to do something like this to resolve the libxml2 problems in the previous release. We will probably start to revisit the specifics week after next.

skef:
@sternenseemann I added an environment variable to the root CMakeLists.txt  that should allow you to force linking against the system LibXML2 on Linux. If you just set it to anything CMake should find it. Give it a shot on your distro and if it doesn't work file an issue against afdko and I'll look at it.

sternenseemann:
@skef thanks that works & we are able to update to 3.9.2 now.

PR #1530, opened 2022-07-04
Ensure `poolCapacity` is at least 1

Heptazhou:
## Description

#1421 introduced only an upper-limit but not a lower one. It causes errors in some cases.

This pr will fix the problem.

## Checklist:


miguelsousa (approved):
Thank you!

Issue #1538, opened 2022-07-14
[mergefonts] Crash on M1


kaydeearts:
Masataka found a bug where `mergefonts` crashes on M1 within his workflow. I am able to reproduce this on a non-M1 Mac with Xcode address sanitizer. Something happens to some glyph file names while mergefonts is running, causing the program to fail when trying to open said glyph file.

I'm currently debugging this — please comment below if you run into a similar issue.

kaydeearts:
This was resolved by #1515, #1523, and #1541.

Issue #1522, opened 2022-06-15
[tx] Replace UFO glif file parsing with libxml2


kaydeearts:
Part of #1468 , #475. Need to replace old parser code for `.glif` files

kaydeearts:
This work was completed as part of https://github.com/adobe-type-tools/afdko/pull/1556

PR #1541, opened 2022-07-18
[uforead] Set currentiFD to 0

kaydeearts:
## Description

I found this while debugging #1538, which looks like the problem was mostly solved for by the libxml2 changes in #1515 and #1523. The only remaining issue seems to be the following:

`currentiFD` is a variable used to track which font dict index is pointed to in the FDArray. This is used in UFO parsing of fontinfo.plist, as well as glyphs files (in that order). While parsing glyphs files, the `currentiFD` is changed to -1. `mergefonts` parses fontinfo.plist again after the glyphs files, which causes a memory leak in non-M1 and a fail in M1 Macs because `currentiFD` should start with 0 (this points to the initial font dictionary).

Since the test doesn't fail on non-M1 unless ASAN is enabled, I did not add a new test.

This is a temporary fix as I'll rewrite glyphs parsing when working on #1522

I also included a small fix where the ufoErrSrcStream was being returned instead of ufoSuccess.

## Checklist:

### DIFF
--- a/c/shared/source/uforead/uforead.c
+++ b/c/shared/source/uforead/uforead.c
@@ -1790,7 +1790,7 @@ static int parseXMLFile(ufoCtx h, char* filename, const char* filetype){
         if (setFontDictKey(h, keyName, cur) && cur != NULL)
            cur = cur->next;
     }
-    return ufoErrSrcStream;
+    return ufoSuccess;
 }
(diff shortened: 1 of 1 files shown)


josh-hadley (approved):
Looks good

Issue #1521, opened 2022-06-15
[tx] Replace UFO content.plist glyphList parsing with libxml2


kaydeearts:
Part of #1468 , #475. Need to replace old parser code for content.plist

kaydeearts:
Done in #1543

PR #1543, opened 2022-07-20
[tx] Replace UFO content.plist/glyphList parsing with libxml2

kaydeearts:
## Description

This PR replaces contents.plist (glyphList) parsing code with libxml2 https://github.com/adobe-type-tools/afdko/issues/1521 as part of the [tx parser replacement with libxml2 project](https://github.com/adobe-type-tools/afdko/projects/7)

I added a comment below on your thoughts on something.

## Checklist:


c/shared/source/uforead/uforead.c
```diff
@@ -1255,23 +1266,17 @@ static int findGLIFRecByName(ufoCtx h, char *glyphName)
...
-        getKeyValue(h, "</string>", state); /* consume the file name */
     } else {
-        char* fileName;
         GLIF_Rec* glifRec;
 
         glifRec = &h->data.glifRecs.array[index];
 
-        fileName = getKeyValue(h, "</string>", state);
         if (fileName == NULL) {
             fatal(h, ufoErrParse, "Encountered glyph reference in alternate layer's contents.plist with an empty file path. Text: '%s'.", getBufferContextPtr(h));
```
kaydeearts, line 1281: Curious to hear your thoughts about this — a key with an empty file path as its value will not fail anymore because the new libxml2 parsing code will simply skip a key that has a missing or NULL value, rather than fail like it did in the older parser code.

I think this is fine but wanted to hear if you have concerns about it. I didn't think of adding warnings since we're trying to reduce the amount of warning messages we have, but I can add them in this case if needed.

skef (dismissed):
I'm not sure I understand every bit of this PR but what I do understand looks good (relative to the one comment).

kaydeearts:
> I'm not sure I understand every bit of this PR but what I do understand looks good (relative to the one comment).

If there's anything I could clarify for you, let me know!

josh-hadley (approved):

PR #1548, opened 2022-07-28
fix reserved ID range in name table

drj11:
[skip ci]

The original text says "IDs 2 and 6 ... are reserved" but in parentheses gives a list of 6 fields that are reserved. "Family, Subfamily, Unique, Full, Version, and FontName" plausibly correspond to the Name IDs 1 through 6 in the OpenType spec https://docs.microsoft.com/en-us/typography/opentype/spec/name#name-ids but annoyingly the OpenType spec doesn't give them canonical short names.

I therefore suspect that "IDs 1 through 6" is intended.

But i haven't cross-checked against what the implementation actually does.

## Checklist:

### DIFF
--- a/docs/OpenTypeFeatureFileSpecification.md
+++ b/docs/OpenTypeFeatureFileSpecification.md
@@ -3371,7 +3371,7 @@ nameid <id> [<string attribute>] <string>;
 ```
 
 An `<id>` is a number specifying the ID of the name string to be added to the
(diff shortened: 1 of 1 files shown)


josh-hadley:
Related discussion: #1457

josh-hadley:
@drj11 after some discussion with the other AFDKO developers, we decided to do some updates to `makeotf` (see the related linked discussion) to make name overriding work better. We'll update the documentation relating to that at the same time.

PR #1556, opened 2022-08-31
[tx] Replace UFO .glif file parsing with libxml2

kaydeearts:
## Description

This PR replaces UFO .glif file parsing code with libxml2 #1522 as part of the [tx parser replacement with libxml2 project](https://github.com/adobe-type-tools/afdko/projects/7). It replaces quite a few parsing functions as well as adds some.

I'll be adding tests soon as part of this PR.

## Checklist:


kaydeearts:
@josh-hadley @skef this is marked as a draft while I add more tests to the PR, but if you'd like to, you can start reviewing whenever available!

Note:I think there will be some refactoring needed in the future, perhaps not necessarily in this PR— I don't like that I had to keep adding global state tracking variables defined at the top of uforead. I would like to make this cleaner. I will still need to track which part of the UFO is being parsed, but I'll manage this better.  Along with this, I'm planning on splitting the setFontDictKey function into a function for each part of the UFO, rather than have all of those keys parsed in one function. I may do all of this refactoring in a different PR, but am open to feedback on this anytime!

Thanks 😄

c/shared/source/uforead/uforead.c
```diff
@@ -1896,69 +1941,38 @@ static int parseGuideline(ufoCtx h, GLIF_Rec* glifRec, int state) {
...
+    while (attr != NULL) {
+        if (xmlAttrEqual(attr, "x"))
+            guideline->x = (float)strtod(getXmlAttrValue(attr), NULL);
+        else if (xmlAttrEqual(attr, "y"))
+            guideline->y = (float)strtod(getXmlAttrValue(attr), NULL);
+        else if (xmlAttrEqual(attr, "angle"))
+            guideline->angle = (float)strtod(getXmlAttrValue(attr), NULL);
+        else if (xmlAttrEqual(attr, "name")) {
+            char* temp = getXmlAttrValue(attr);
+            size_t nameLen = strlen(temp) + 1;
```
skef, line 1954: `copyStr()` for these too?

c/shared/source/uforead/uforead.c
```diff
@@ -2041,235 +2027,217 @@ static int parseNote(ufoCtx h, GLIF_Rec* glifRec, int state) {
...
 }
 
 static int parseGLIF(ufoCtx h, abfGlyphInfo* gi, abfGlyphCallbacks* glyph_cb, Transform* transform);
 
-static int parseComponent(ufoCtx h, GLIF_Rec* glifRec, abfGlyphCallbacks* glyph_cb, int state, Transform* transform) {
+
+static int setParseXMLComponentValue(ufoCtx h, abfGlyphInfo* gi, abfGlyphCallbacks* glyph_cb, GLIF_Rec* glifRec, Transform* transform, Transform localTransform, Transform *newTransform, int result) {
+    Transform concatTransform;
+    if (gi == NULL) {
+        fprintf(stderr, "Missing component base attribute. Glyph: %s, Context: %s.\n", glifRec->glifFilePath, getBufferContextPtr(h));
```
skef, line 2123: Why the out-of-band-warning instead of calling `message()`?

c/shared/source/uforead/uforead.c
```diff
@@ -3237,212 +3207,18 @@ static int parseGLIF(ufoCtx h, abfGlyphInfo* gi, abfGlyphCallbacks* glyph_cb, Tr
     h->cb.stm.clientFileName = glifRec->glifFilePath;
     h->stm.src = h->cb.stm.open(&h->cb.stm, UFO_SRC_STREAM_ID, 0);
 
-    /* Note that in preParseGlyph, we found the start of the <outline> element, and set gi->sup.begin
-    to that, so we can seek to this and skip all the preceding stuff, if any.
-    */
     if (h->stm.src == NULL || h->cb.stm.seek(&h->cb.stm, h->stm.src, gi->sup.begin)) {
         fprintf(stderr, "Failed to open glif file in parseGLIFOutline. Glyph: %s. Context: %s\n.", glifRec->glyphName, getBufferContextPtr(h));
```
skef, line 3211: I guess these crept in somehow.

c/shared/source/uforead/uforead.c
```diff
@@ -1001,6 +1016,62 @@ static void setFontMatrix(ufoCtx h, abfFontMatrix* fontMatrix, int numElements)
...
+    int stemFlags = 0;
+    int count = 0;
+    int isH = !(stemFlags & ABF_VERT_STEM);
+//    if ((transform != NULL) && (!transform->isOffsetOnly)) { looks like transform is ALWAYS NULL
+//        /* We omit stems if the stems are being skewed */
+//        if (isH && (transform->mtx[2] != 0.0))
+//            return result;
+//        if ((!isH) && (transform->mtx[1] != 0.0))
+//            return result;
+//    }
```
josh-hadley, line 1045: Remove WIP(?) commented-out code (or check/fix re: `transform`)

tests/makeotf_test.py
```diff
@@ -613,18 +613,21 @@ def test_outline_from_processed_layer_bug703():
...
-
-    stderr_path = runner(
-        CMD + ['-s', '-e', '-o',
-               'f', f'_{get_input_path(input_filename)}',
-               'o', f'_{otf_path}'])
-
-    with open(stderr_path, 'rb') as f:
-        output = f.read()
-    assert b"unhandled token: <foo" in output
+# We've decided to reduce verbosity until we add verbose flags.
```
josh-hadley, line 616: Maybe mark this with `# TODO` to make it stand out for future adjustments

josh-hadley (changes requested):
minor twiddles

c/shared/source/uforead/uforead.c
kaydeearts: I actually put these back just in case. Now I keep transform within the parseState struct in ufoCtx, so it's accessible anytime.

c/shared/source/uforead/uforead.c
```diff
@@ -1738,22 +1645,110 @@ static int parseXMLFile(ufoCtx h, char* filename, const char* filetype){
...
+    cur = cur->xmlChildrenNode;
+    if (xmlKeyEqual(cur, "dict") && xmlKeyEqual(cur->children, "key")) {
+        cur = cur->children;
+        while (cur != NULL) {
+            char* keyName = parseXMLKeyName(h, cur);
+            cur = cur->next;
+            if (h->parseState.UFOFile == preParsingGLIF) {     /* when called from preParseGLIF */
+                if (strEqual(keyName, "com.adobe.type.cid.CID")) {
+                    keyValue = parseXMLKeyValue(h, cur);
+                    h->parseState.GLIFInfo.currentCID = atoi(keyValue);
```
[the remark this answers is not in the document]
kaydeearts: Good catch! I just created a test case for this and it crashes. Would you suggest I add the NULL check to right before setting h->parseState.GLIFInfo.currentCID, or should I return a value like -1 instead of NULL in parseXMLKeyValue so that it fixes this scenario for other callers?

I also realized that if the value of com.adobe.type.cid.CID is a blank value (like the example below) then com.adobe.type.cid.CID is set to 0. I feel like it also probably needs to remain -1 (which is currently what it starts with when unset). 

```
      <key>com.adobe.type.cid.CID</key>
      <integer></integer>
```


I'll look for more places where NULL may be returned from parseXMLKeyValue and the other functions. Thanks!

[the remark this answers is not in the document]
skef: I think being more explicit with NULL checks is probably better for this. Returning "fake" values can eventually result in those values being used.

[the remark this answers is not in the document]
skef: (Although I'd have to look in more detail to give you a conclusive answer ...)

c/shared/source/uforead/uforead.c
```diff
@@ -1682,32 +1681,39 @@ static char* getXmlAttrValue(xmlAttr *attr){
     return (char*) attr->children->content;
 }
 
-static void parseXMLLib(ufoCtx h, xmlNodePtr cur) {
+static bool setXMLLib(ufoCtx h, xmlNodePtr cur, char* keyName) {
```
skef, line 1684: You've declared this `bool`, presumably to track whether you ran into some problem parsing (given the check on the call), but you're not returning anything.

c/shared/source/uforead/uforead.c
```diff
@@ -1682,32 +1681,39 @@ static char* getXmlAttrValue(xmlAttr *attr){
...
+                h->parseState.GLIFInfo.currentCID = atoi(keyValue);
+                h->top.sup.flags |= ABF_CID_FONT;
+                h->top.sup.srcFontType = abfSrcFontTypeUFOCID;
+            } else if (strEqual(keyName, "com.adobe.type.cid.iFD"))
+                h->parseState.GLIFInfo.currentiFD = atoi(keyValue);
+        }
+    } else if (h->parseState.UFOFile == parsingGLIF) {  /* when called from parseGLIF */
+        if (strEqual(keyName, "com.adobe.type.autohint.v2")) {
+            h->parseState.type1HintDataV2 = true;
+            if (cur != NULL)
```
skef, line 1700: If `cur` is NULL, do we want to just ignore the fact there was a recognized key but no value here? Seems like that should be a warning at least.

c/shared/source/uforead/uforead.c
```diff
@@ -2475,9 +2481,30 @@ static void setFlexListArrayValue(ufoCtx h) {
     freeValueArray(h);
 }
 
+static bool setType1HintDataV2(ufoCtx h, char* keyName, HintMask* curHintMask, xmlNodePtr cur) {
+    char* keyValue;
+    if (strEqual(keyName, "hintSetList")) {
+        parseHintSetListV2(h, cur);
```
skef, line 2487: Are all of the uses of `cur` in this method NULL-safe? Seems like since you parsed a key at the level above already, it should always have a value, it would be easier to check whether `cur` is NULL above, deal with that condition there, and always be sure `cur` is not NULL here.

c/shared/source/uforead/uforead.c
kaydeearts: For now, we've been aiming to reduce the amount of warnings until we add the verbosity flag option. However, I'm curious what @frankrolf thinks about this specifically. Frank, if within a UFO's .glif file's <lib> there is an empty `com.adobe.type.autohint.v2`, would it be helpful to the user to receive a warning?

skef: My experimental repo branch (not yet copied back from my machine) has already pulled all the C-code-level logging messages into a single framework with multiple "levels". So after we start working with that code it will be easier to unify the logging across all of our tools.

kaydeearts: Sounds good! So we can keep the warning messages for when we work with your changes

kaydeearts:
@skef @josh-hadley This PR is up to date with the latest changes you both requested, and I've added more tests covering edge cases & etc. Let me know if you have any more thoughts on this!

josh-hadley (approved):

skef (approved):
At some point we should nail down what the return value of `stm_xml_read` represents, as it's sort of being used as a weird boolean right now, but I don't see that as a reason to hold things up.

PR #1561, opened 2022-09-18
Boost number of makeinstancesufo pool processors by one

skef:
## Description

The makeinstancesufo tests currently fail on single-processor systems due to requesting 0 cores for the multiprocessing pool. This bumps the count by one.

The motivation to leave one processor free is understandable. However, if the "organizing" thread isn't very active when the pool is running it seems fine to rely on the timesharing in the OS and just use all the cores. 

## Checklist:


josh-hadley (approved):

josh-hadley:
I approved this and OK'd @skef to bypass the check requirements because of some weirdness with re-triggering the required actions. They ran and passed on the skef/afdko branch so I'm satisfied that everything is 🆗

Issue #1539, opened 2022-07-15
[tx] Add iFD to glyphs mappings to groups.plist


kaydeearts:
Within groups.plist for CID-keyed UFO, add a dictionary of iFDs that point to a list of glyphs to identify which glyphs use which fdicts within the FDArray

punchcutter:
I updated the test file tests/tx_data/input/cid_roundtrip/testCID.ufo with the new setup for FDArraySelect groups in groups.plist. Removed PrivateDicts from the FDArrays in lib.plist and moved the data in them to the top level. Also changed `com.adobe.type.postscriptCIDMap` to only map glyph names to CID instead of having iFD in there as well.

kaydeearts:
Implemented in #1576 . Here's an example of [what the groups.plist should look like!](https://github.com/adobe-type-tools/afdko/blob/develop/tests/tx_data/input/cid_roundtrip/testCID.ufo/groups.plist)

PR #1576, opened 2022-11-15
[tx][libxml2-improved] Move FDArray from fontinfo.plist to lib.plist

### LINKED ISSUE #1542: [tx] Move FDArray from fontinfo.plist to lib.plist — libxml2 version
This was done in #1380, but will be redone as a part of the new libxml2 changes.

kaydeearts:
## Description

Fixes #1542 and #1539
Does what #1380 intended, but rewritten for the new libxml2 improvements + other considerations.
🚀 

## To-Dos:
- [x] ufowrite: write FDArray in lib.plist instead of fontinfo.plist
- [x] uforead: parse FDArray from lib.plist as well as fontinfo.plist (to support old & new files)
- [x] .glif files: current solution is that iFD & CID is held in a CIDMap in lib.plist. Is this what we want?
- [x] update tests to reflect the changes
- [x] Fix test failures
- [x] Remove support for FDArray in fontinfo.plist
- [x] Edit CIDMap to only have cids
- [x] Move iFD information into a group in groups.plist
- [x] Add tests for groups.plist parsing
- [x] Add tests for CIDMap in lib.plist parsing
- [x] Fix Windows heap corruption fails

## Checklist:


kaydeearts:
Question (mainly for @skef ): Last year, we had written this so that the cid and iFD mappings would be moved to a CIDMap in lib.plist. We have discussed moving the iFD mappings to groups.plist in #1539, so I'm wondering if we should keep the CIDMap as well as have the iFD group, or move the CIDMap to groups.plist, or have only one of them? 

For now, I kept the CIDMap in lib.plist in this PR as I mainly focused on refactoring my PR from last year. 
I can now make changes as needed since tests pass.

skef:
The CIDmap isn't a group so it can't/shouldn't go in groups.plist.

I don't see a problem with separating the two (keeping the map in lib.plist and having the group in groups.plist). Some LGC fonts will have FD mappings but no CIDmap.

kaydeearts:
Question about the key naming: Shall we discuss if we want keys to have "com.adobe.type" or not? Currently, the code should support keys that do and don't have it (since what we currently have in the FDK are the keys without the `com.adobe.type` in them, and I didn't want to break older files.

skef:
Was the previous code ever really in production?

kaydeearts:
> Was the previous code ever really in production?

@skef No, it only ever stayed in PR #1380 and we never ended up merging. Then the plan became to rewrite it after the libxml2 updates. However, Zachary used that branch to release Source Han Serif last year and still uses it for font development.

punchcutter:
I already started updating to use the new way. No reason at all to keep the old way. iFD should go into groups.plist and CIDs in the CIDMap in lib.plist without all the com.adobe stuff because that's irrelevant. The top level com.adobe.postscriptCIDMap or whatever that name is (writing on my phone) has com.adobe and that should just be a dict with names mapped to CID. The only reason it had com.adobe or even "CID" was because I originally made it a dict with CID and iFD. Since we are moving iFD nothing needs to be named iFD or CID.

c/shared/source/uforead/uforead.c
```diff
@@ -1144,46 +1158,69 @@ static void addGLIFRec(ufoCtx h, char* glyphName, xmlNodePtr cur) {
...
     if (fileName == NULL) {
         fatal(h, ufoErrParse, "Encountered glyph reference in contents.plist with an empty file path. Text: '%s'.", getBufferContextPtr(h));
     }
-    newGLIFRec->glifFileName = memNew(h, 1 + strlen(fileName));
-    sprintf(newGLIFRec->glifFileName, "%s", fileName);
-    newGLIFRec->altLayerGlifFileName = NULL;
+    int nameLength = strlen(fileName) - 5;
+    char *subbuff = memNew(h, (sizeof(char*) * nameLength));
+    memcpy(subbuff, &fileName[0], nameLength);
+    subbuff[nameLength] = '\0';
```
skef, line 1181: You didn't allocate this byte.

skef (commented):
I got a bit into this but I think it would be more useful for me to wait until the older conventions are removed from the code.

punchcutter:
I tried the latest kd-fdarray-libplist-libxml2-2 and found two problems:

1. Running tx fails on a UFO if lib.plist has `public.glyphOrder` after the CID information. `public.glyphOrder` needs to be defined first in the lib.plist, but that doesn't make any sense. I thought the plan was to read in everything in whatever order since nothing in any of these XMLs is ordered. We can't expect keys to be in any particular order.   
2. All keys with `com.adobe.type.cid` should be changed to `com.adobe.type`. That's my fault for putting `cid` in there in the first place, but it should be removed.

punchcutter:
Something seems off with CID order defined in com.adobe.type.postscriptCIDMap vs public.glyphOrder. I changed the order in public.glyphOrder from 
```
    <key>public.glyphOrder</key>
    <array>
      <string>cid00000</string>
      <string>cid00001</string>
```
to
```
    <key>public.glyphOrder</key>
    <array>
      <string>cid00001</string>
      <string>cid00000</string>
```

and that changed the CID assigned. The glyph order _could_ be the same, but there's nothing about public.glyphOrder that should dictate the CID mapping which is defined in com.adobe.type.postscriptCIDMap.

punchcutter:
If there is a mismatch between the number of FDArrays defined in lib.plist and the number of FDSelect groups in groups.plist we should error out. Right now if we drop an FDArray from `com.adobe.type.postscriptFDArray` then we still get an empty one created with no warning. We want to be sure these are actually defined so we should fail like we do if a glyph isn't assigned to a dict in groups.plist.

punchcutter:
Kind of related to my last comment running tx alone on a UFO outputs an extra empty FontDict. Running `tx -t1` and outputting to a file is fine and the file doesn't have an extra dict. More of a cosmetic thing I think. 

```
## FontDict[18]
## Private
BlueScale           0
BlueShift           0
BlueFuzz            0
StdHW               0
StdVW               0
ExpansionFactor     0

```

c/shared/source/uforead/uforead.c
```diff
@@ -1813,10 +1903,70 @@ static bool parseFontInfoFDArray(ufoCtx h, xmlNodePtr cur) {
...
+        }
+        cur = cur->next;
+    }
+    memFree(h, FDIndex);
+    return true;
+}
+
+static bool parseCIDMap(ufoCtx h, xmlNodePtr cur) {
+    h->parseState.CIDMap = true;
+    h->top.sup.flags |= ABF_CID_FONT;
```
punchcutter, line 1938: I'm not sure we should be setting ABF_CID_FONT here because you can delete the other required keys and if `com.adobe.type.postscriptCIDMap` still exists the UFO will be read as CID-keyed and then fail because not everything required is set. I think it might make sense to have a preparser to check for all required keys first. We only need to read top level keys and look for name, ROS, CIDMap, and FDArrays. If any of those are missing we should bail on trying to read as CID-keyed. If we do that then this line is fine because we can set the flag if CIDMap comes before name and ROS.

kaydeearts:
> If there is a mismatch between the number of FDArrays defined in lib.plist and the number of FDSelect groups in groups.plist we should error out. Right now if we drop an FDArray from `com.adobe.type.postscriptFDArray` then we still get an empty one created with no warning. We want to be sure these are actually defined so we should fail like we do if a glyph isn't assigned to a dict in groups.plist.

My latest changes now implement this. If there aren't enough FDArrays as FDSelect groups defined, it will error out.
I was also going to check to see if the assigned FDArray's name matches the assigned FDSelect group, but I remembered that we had left that naming portion as optional/just for the user, so avoided doing that. Let me know if that's something we do want!

kaydeearts:
> Kind of related to my last comment running tx alone on a UFO outputs an extra empty FontDict. Running `tx -t1` and outputting to a file is fine and the file doesn't have an extra dict. More of a cosmetic thing I think.
> 
> ```
> ## FontDict[18]
> ## Private
> BlueScale           0
> BlueShift           0
> BlueFuzz            0
> StdHW               0
> StdVW               0
> ExpansionFactor     0
> ```

Directly out of the most recent changes, parsing a UFO with mismatched FDArrays/FDSelect groups with or without the `-t1` arg will now error out.

c/shared/source/uforead/uforead.c
kaydeearts: My latest commit implements these checks, including FDArray. If we'd like to remove some keys from being required, I can just comment those out later.

Do we want to error/warn the user if there are missing required keys but the font looks like it should be CID-keyed? (like if one or more CID-keyed keys are present, but not all of them). I'll add tests for this once we make a decision on this.

c/shared/source/uforead/uforead.c
```diff
@@ -1815,6 +1834,44 @@ static int parseXMLGlifFile(ufoCtx h, xmlNodePtr cur, int tag, unsigned long *un
...
+        return false;
+    else if (h->requiredCIDKeys.Supplement == false)
+        return false;
+    else if (h->requiredCIDKeys.postscriptFDArray == false)
+        return false;
+    else if (h->requiredCIDKeys.postscriptCIDMap == false)
+        return false;
+    else
+        return true;
+}
```
kaydeearts, line 1873: @skef I'm curious to hear what you think about this part of my code in particular whenever you have time — I thought about it and this was the best/only way I could think of checking for required keys within C. If you have a better/more efficient idea for implementation, please let me know! 😄

skef: I don't think this is a big deal either way, but if you wanted to reduce the second part you could turn requiredCIDKeys into a uint16_t (or whatever), make a bit flags enum for each of the keys, and set those flags with `|=` in the top part. Then at the end you just need to verify that all the bits are set, probably by having an extra enum value representing the max, and just return `requiredCIDKeys == (maxBit - 1)`.

kaydeearts: @skef thanks for the recommendation! I liked it much more than the way I had it. My latest commit implements it the way you described.

punchcutter:
I tried the latest and if I remove com.adobe.type.postscriptCIDMap it still reads as CID-keyed with all glyphs set to CID 0.

c/shared/source/uforead/uforead.c
```diff
@@ -1813,10 +1951,72 @@ static bool parseFontInfoFDArray(ufoCtx h, xmlNodePtr cur) {
...
+/* set each GLIF_REC's iFD based on the FDArraySelect groups*/
+static bool parseFDArraySelectGroup(ufoCtx h, char* FDArraySelectKeyName, xmlNodePtr cur) {
+    GLIF_Rec *g;
+
+    if (!xmlStrEqual((cur)->name, (const xmlChar *) "array"))
+        return false;
+
+    char* FDIndex = memNew(h, sizeof(char) * 2);
+    strncpy(FDIndex, FDArraySelectKeyName + 14, 2);
+    if (FDIndex[1] == '.')
```
skef, line 1963: Sadly, I'm pretty sure one can have many more than 10 elements in an FDArray. I'm pretty sure I've run into actual fonts with more than that.

One of the `strtol` variants is probably what you want here.

c/shared/source/uforead/uforead.c
```diff
@@ -1813,10 +1951,72 @@ static bool parseFontInfoFDArray(ufoCtx h, xmlNodePtr cur) {
...
+        if (keyName != NULL) {
+            g = addNewGLIFRec(h, keyName, NULL);
+            if (g == NULL)
+                continue;
+            cur = cur->next;
+            if (cur == NULL)
+                continue;
+            char* cid = parseXMLKeyValue(h, cur);
+            if (cid != NULL)
+                g->cid = atoi(cid);
```
skef, line 2007: This is probably OK as garbage will result in a 0 value that will be detected later, but it's better to sanity-check whether a 0 value makes sense. (Can we tell at this point whether a glyph is a .notdef?)

c/shared/source/ufowrite/ufowrite.c
```diff
@@ -650,6 +553,219 @@ static void writeBlueValues(ufwCtx h, abfPrivateDict *privateDict){
...
+    if (h->top->sup.flags & ABF_CID_FONT) {
+        if (h->top->cid.CIDFontName.ptr != NULL) {
+            writeLine(h, "\t<key>com.adobe.type.CIDFontName</key>");
+            sprintf(buffer, "\t<string>%s</string>", h->top->cid.CIDFontName.ptr);
+            writeLine(h, buffer);
+            writeLine(h, "\t<key>com.adobe.type.ROS</key>");
+            sprintf(buffer, "\t<string>%s-%s-%ld</string>", h->top->cid.Registry.ptr, h->top->cid.Ordering.ptr, h->top->cid.Supplement);
+            writeLine(h, buffer);
+        }
+        writeFDArray(h, h->top, buffer);
```
skef, line 659: What if there's an FDArray in a non-CID font?

c/shared/source/ufowrite/ufowrite.c
```diff
@@ -650,6 +553,219 @@ static void writeBlueValues(ufwCtx h, abfPrivateDict *privateDict){
...
+    for (i = 0; i < h->glyphs.cnt; i++) {
+        sprintf(buffer, "\t\t<key>%s</key>", h->glyphs.array[i].glyphName);
+        writeLine(h, buffer);
+        sprintf(buffer, "\t\t<integer>%d</integer>", h->glyphs.array[i].cid);
+        writeLine(h, buffer);
+    }
+    writeLine(h, "\t</dict>");
+}
+
+static void writeGlyphOrder(ufwCtx h) {
```
skef, line 618: Seems like this does more than just write the glyph order.

c/shared/source/ufowrite/ufowrite.c
```diff
@@ -650,6 +553,219 @@ static void writeBlueValues(ufwCtx h, abfPrivateDict *privateDict){
...
+            writeLine(h, buffer);
+        }
+        writeFDArray(h, h->top, buffer);
+        writeCIDMap(h, h->top, buffer);
+    }
+    writeLine(h, "</dict>");
+    writeLine(h, "</plist>");
+
+    /* Close dst stream */
+    flushBuf(h);
```
skef, line 666: I'm worried we may have missed a significant aspect of this implementation. 

One of the reasons we chose to move these structures from `fontinfo.plist` to `lib.plist` is that the keys in the former are fixed and implementations throw other keys away. If we want our custom data to stay we need to use `lib.plist` (and `groups.plist`) instead. 

However, if someone is using `tx` to take a round-trip from UFO to UFO we need to honor the same convention: if there's a key in `lib.plist` that isn't one of ours it needs to be preserved. That probably means opening the file with libxml2, modifying that DOM, and then outputting it via libxml2 as well. If the file isn't there we can have this writeLine based stuff as an alternative or (probably preferable) start with a blank LibXML2 DOM object and add our elements to it.

Trying to go through our custom stream IO crap (I really hate it) to do that is probably not worth it, especially given that we're opening new files here anyway, so I would just do the most natural LibXML2-flavored thing.

c/shared/source/ufowrite/ufowrite.c
```diff
@@ -650,6 +553,219 @@ static void writeBlueValues(ufwCtx h, abfPrivateDict *privateDict){
...
+    writeLine(h, PLIST_DTD_HEADER);
+    writeLine(h, "<plist version=\"1.0\">");
+    writeLine(h, "<dict>");
+    writeFDArraySelect(h, top, buffer);
+    writeLine(h, "</dict>");
+    writeLine(h, "</plist>");
+
+    /* Close dst stream */
+    flushBuf(h);
+    h->cb.stm.close(&h->cb.stm, h->stm.dst);
```
skef, line 724: Same comment for `groups.plist` as the `libs.plist` one above.

c/shared/source/ufowrite/ufowrite.c
```diff
@@ -650,6 +553,219 @@ static void writeBlueValues(ufwCtx h, abfPrivateDict *privateDict){
...
+    writeLine(h, "\t<key>creator</key>");
+    writeLine(h, "\t<string>com.adobe.type.tx</string>");
+    writeLine(h, "\t<key>formatVersion</key>");
+    writeLine(h, "\t<integer>2</integer>");
+    writeLine(h, "</dict>");
+    writeLine(h, "</plist>");
+
+    /* Close dst stream */
+    flushBuf(h);
+    h->cb.stm.close(&h->cb.stm, h->stm.dst);
```
skef, line 761: And same thing here too, probably.

skef:
I could be wrong about the UFO `.plist` preservation question. I suggest consulting with Josh and Zachary about that.

c/shared/source/ufowrite/ufowrite.c
punchcutter: Yes, we want to preserve existing data in a UFO when possible. If we go UFO > t1 then we obviously can't round trip all the data because there is nowhere to store it. But if we want to, for example, subset a CID-keyed UFO then anything in lib.plist not ours should be retained while our keys should be updated. Doing a basic `tx -ufo -o newfile.ufo oldfile.ufo` should preserve everything because nothing else has happened. Something like `tx -ufo -o newfilesubset.ufo oldfile.ufo -g 0-44` should also remove everything except these glyphs from the UFO, public.glyphOrder, CIDMap, and groups.plist.

c/shared/source/uforead/uforead.c
punchcutter: If cid == 0 then it is the .notdef glyph so we shouldn't ever get 0 except for this one.

skef: @punchcutter `atoi()` returns 0 when its fed a string that doesn't contain a number, so if we can't verify here that we *should* be getting zero when we actually get zero we have a data sanitizing problem.

punchcutter: We can hope the associated glyph name is either cid00000 or .notdef, but we can't guarantee it so we need some other way to verify that zero means zero.

skef: Probably best to switch to strtol/strtoul everywhere and check the `errno` value when it returns 0.

c/shared/source/uforead/uforead.c
```diff
@@ -1132,58 +1162,83 @@ static long getGlyphOrderIndex(ufoCtx h, char* glyphName) {
...
 }
 
 static void updateGLIFRec(ufoCtx h, char* glyphName, xmlNodePtr cur) {
-    int index;
+    GLIF_Rec* glifRec;
     char* fileName = parseXMLKeyValue(h, cur);
     if (!keyValueParseable(h, cur, fileName, glyphName))
         return;
+    if (fileName == NULL) {
+        /* this is basically muted for now, as the previous check will return and skip if not parseable.
```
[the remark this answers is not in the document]
kaydeearts: I agree it sounds weird — I think my intention was to say that with a verbosity flag, the user would be told that the GLIF is missing. I'll downgrade this to a warning message instead, since in the current implementation we just ignore if this happens.

c/shared/source/uforead/uforead.c
```diff
@@ -1332,17 +1412,17 @@ static void addCharFromGLIF(ufoCtx h, int tag, char* glyphName, long char_begin,
         chr->flags = 0;
         chr->tag = tag;
         chr->iFD = 0;
-        if (h->parseState.GLIFInfo.currentCID >= 0) {
-            chr->cid = h->parseState.GLIFInfo.currentCID;
-            if (h->parseState.GLIFInfo.currentiFD < 0){
-                    fatal(h, ufoErrParse, "Warning: glyph '%s' missing FDArray index within <lib> dict", glyphName);
+        if (glifRec->cid >= 0) {
```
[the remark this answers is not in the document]
kaydeearts: I'm leaving adding support in for non-cid-keyed fonts for another PR after this one, does that sound fine?

c/shared/source/uforead/uforead.c
```diff
@@ -1747,35 +1826,94 @@ static int parseXMLGlifFile(ufoCtx h, xmlNodePtr cur, int tag, unsigned long *un
...
+    else if (strEqual(keyName, "com.adobe.type.CIDFontName"))
         top->cid.CIDFontName.ptr = copyStr(h, keyValue);
-    else if (strEqual(keyName, "com.adobe.type.cid.Registry"))
-        top->cid.Registry.ptr = copyStr(h, keyValue);
-    else if (strEqual(keyName, "com.adobe.type.cid.Ordering"))
-        top->cid.Ordering.ptr = copyStr(h, keyValue);
-    else if (strEqual(keyName, "com.adobe.type.cid.Supplement"))
-        top->cid.Supplement = atol(keyValue);
+    else if (strEqual(keyName, "com.adobe.type.ROS")) {
+        value = strtok_r(keyValue, "-", &ROSValues);
```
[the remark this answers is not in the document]
kaydeearts: I see, thanks for the catch. @punchcutter since we require this key, I assume that if a value is NULL then we'd like to fail/warn? What do you think?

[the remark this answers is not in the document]
punchcutter: If the ROS key exists then we should assume this is meant to be a CID-keyed font and fail if the ROS value is NULL. We should also make sure that the ROS value conforms to `string-string-integer` where the strings can only contain alphanumerics and underscores.

c/shared/source/uforead/uforead.c
```diff
@@ -1747,35 +1826,94 @@ static int parseXMLGlifFile(ufoCtx h, xmlNodePtr cur, int tag, unsigned long *un
...
+        top->cid.Registry.ptr = copyStr(h, value);
+        value = strtok_r(ROSValues, "-", &ROSValues);
+        top->cid.Ordering.ptr = copyStr(h, value);
+        value = strtok_r(ROSValues, "-", &ROSValues);
+        top->cid.Supplement = atol(value);
+    } else if (setFontInfoFD(h, keyName, keyValue))
+        return true;
+    else if (setFontInfoPD(h, keyName, keyValue))
```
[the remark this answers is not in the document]
kaydeearts: To clarify, which two checks are you asking about? If you''re referring to `setFontInfoFD` and `setFontInfoPD`:
`setFontInfoFD` is for each font dict (I should rename this to setFontInfoDict for clarity)
`setFontInfoPD` is for each private dict within each font dict (I should rename this to setFontInfoPrivateDict for clarity).

Or is it something else you're referring to?

c/shared/source/uforead/uforead.c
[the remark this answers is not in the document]
skef: Yes, that's fine.

c/shared/source/uforead/uforead.c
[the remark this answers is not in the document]
skef: Oh, yeah, I didn't see the different letter. Sorry about that.

c/shared/source/uforead/uforead.c
kaydeearts: Not sure why I even wrote it assuming it would have less than 10...should be fixed now + added a new test for roundtripping with 100 FDArrays

c/shared/source/ufowrite/ufowrite.c
kaydeearts: renamed the parse & write funcs in uforead & ufowrite from GlyphOrder -> LibPlist

c/shared/source/ufowrite/ufowrite.c
kaydeearts: [Similarly to the comment here](https://github.com/adobe-type-tools/afdko/pull/1576#discussion_r1109234784), I think we can support FDArrays in non-CID fonts in an upcoming PR — let me know if you have thoughts!

c/shared/source/uforead/uforead.c
kaydeearts: @skef latest commit replaces strtol/strtoul/strtod with strtolCheck/strtoulCheck/strtodCheck functions

skef (dismissed):
I only reviewed the changes since my last review because I remember it being in good shape otherwise. So while I think you should probably get a separate sign-off from @punchcutter I'm approving this from my end.

c/shared/source/ufowrite/ufowrite.c
kaydeearts: An implementation for all UFO keys to be preserved (including unidentified keys) for UFO -> UFO roundtripping will come in a future PR made after this one merges.

punchcutter:
The one big thing that I see right now is that going `tx -ufo -o test.ufo test.pfa` doesn't create a groups.plist so any further tx'ing on the UFO fails 
```
tx: (ufr) Warning: FDArraySelect not defined for cid-keyed font
tx: (ufr) glyph 'cid00000' missing FDArray index within <lib> dict
```

We _could_ merge this and then address that. That second message is also unclear because lib.plist _does_ have FDArray, but what's actually missing is groups.plist with an FDArraySelect containing that glyph.

punchcutter (approved):
LGTM. Let's merge it!

PR #1591, opened 2023-01-04
Always use blend for stem storage in cffwrite_t2cstr.c

skef:
## Description

The current version of this file stores stems in a union, with just a float for edge0 and edge1 in one case and a blend structure in the other. CFF2 uses the blend while CFF uses just the float. The blend `.value` happens to coincide with the float.

This sort of works but the model breaks down for non-blended stems in CFF2 output. It's better to just use the blend structure for all cases with hasBlend set to false if there's no blend data. That way CFF output can ignore everything but the "default" value and CFF2 output can correctly detect whether it needs to output a blend.

## Checklist:

### DIFF
--- a/c/shared/source/cffwrite/cffwrite_t2cstr.c
+++ b/c/shared/source/cffwrite/cffwrite_t2cstr.c
@@ -29,14 +29,8 @@
 
 typedef struct /* Stem hint */
 {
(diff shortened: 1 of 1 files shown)


cjchapman (approved):
👍

PR #1601, opened 2023-01-20
Stop forcing static LibXML2 build on Linux

skef:
## Checklist:


.github/workflows/build_wheels.yml
```diff
@@ -57,8 +57,10 @@ jobs:
         CIBW_ARCHS_LINUX: x86_64
         CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
         CIBW_SKIP: "*musllinux*"
-        CIBW_BEFORE_ALL_LINUX: "yum install -y libuuid-devel && yum install -y libxml2-devel"
-        CIBW_ENVIRONMENT_LINUX: "CFLAGS='-I/usr/include/libxml2'"
+        # CIBW_BEFORE_ALL_LINUX: "yum install -y libuuid-devel && yum install -y libxml2-devel"
+        # CIBW_ENVIRONMENT_LINUX: "CFLAGS='-I/usr/include/libxml2'"
```
kaydeearts, line 61: Just a question: are we keeping this uncommented so users can force a static installation if needed, or for our own references in the future?

skef: The first thing I would say is that the github workflows are pretty much entirely "ours" -- the contents of these files have no impact on "downstream" users, except perhaps in the sense that users who clone the repo on github itself might activate workflows (but probably not this one, since I don't think we trigger it ourselves except at release time).

I suppose I kept the commented lines as a reference for how to build a wheel on "GitHub-world" Linux dynamically linked to LibXML2. You tracked that info down at some point, and even if we've decided it doesn't work it may be useful down the road.

As for comments my view is that because the lines below include setting `FORCE_BUILD_LIBXML2` it's pretty clear what's going on, but YMMV.

.flake8
```diff
@@ -5,7 +5,7 @@
 select = E,F,W,B,B901,B902,B903
 
 # ignore W504 line break before binary operator
-ignore = W504
```
josh-hadley, line 8: This is a new flake8 warning and adding it to ignore is fine for now but reading it I think we should probably make some efforts to update the test code (per the flake8 warning: be more specific than `pytest.raises(Exception)` – use the specific Exception that is expected to be raised) and remove B017 from the ignore list once that's done.

skef: yeah, I added this because it suddenly appeared and was causing the tests to fail. I'm fine with fixing it when we have time.

kaydeearts (approved):
Looks good to me with your comments on my question + agreement on fixing flake8 errors once we have time in the future (but we should make a separate issue for that)

PR #1615, opened 2023-03-01
Fix GOADB order of processing multiple unicode assignments

### LINKED ISSUE #1273: for glyphs named uniXXXX, is makeotf following the documented GOADB behavior?
According [to the docs](http://adobe-type-tools.github.io/afdko/MakeOTFUserGuide.html#glyphorderandaliasdb-goadb):

> a) If the third field of the GOADB record for a glyph contains a Unicode value in the form uniXXXX or uXXXX\[XX\] (see [note](#unicode_note)), assign that Unicode value to the glyph. **Else** b);

> b) If a glyph name is in the [Adobe Glyph List For New Fonts](https://github.com/adobe-type-tools/agl-aglfn/blob/master/aglfn.txt), use the assigned Unicode value. **Else** c);

> c) If the glyph name is in the form uniXXXX or uXXXX\[XX\] (see [note](#unicode_note)), assign the Unicode value. **Else** d);

> d) Do not assign any Unicode value.

The use of “else” in the rules above suggests that if the conditions of that branch are met, then the behavior is triggered, and the heuristic descends no further. 

In particular, if a glyph is named `uniXXXX` and has an explicit codepoint provided in the third column of its GOADB record, then the **first** branch of this rule should be triggered: the explicit codepoint should supersede the codepoint-inference rule in the third branch.

But this doesn’t seem to be what happens. Rather, if a glyph is named `uniXXXX`, it gets a codepoint of `XXXX`, even if the GOADB says otherwise.

Let’s have a test case! Attached is a PFB with one glyph called `uni2032` and a GOADB asking that it receive PUA codepoint `uE00D4`:

```
 uni2032 uni2032 uE00D4

```

(Don’t forget the blank line at the end! `makeotf` will crash without it!)

We expect that the glyph named `uni2032` should get the codepoint `uE00D4`. I generate the font like so:

```
makeotf -f unicode-test.pfb -gf GOADB
```

By the way, the [docs also say](http://adobe-type-tools.github.io/afdko/MakeOTFUserGuide.html#makeotf-options):

> the –r or –ga options are NOT specified, the effect is to use the Unicode assignments from the third column of the GOADB without renaming the glyphs.

Yes! We want that codepoint from the third column.

But here’s what we see in the `cmap` of the generated OTF:

```
<?xml version="1.0" encoding="UTF-8"?>
<ttFont sfntVersion="OTTO" ttLibVersion="3.6">

  <cmap>
    <tableVersion version="0"/>
    <cmap_format_4 platformID="0" platEncID="3" language="0">
      <map code="0x2032" name="uni2032"/><!-- PRIME -->
    </cmap_format_4>
    <cmap_format_6 platformID="1" platEncID="0" language="0">
    </cmap_format_6>
    <cmap_format_4 platformID="3" platEncID="1" language="0">
      <map code="0x2032" name="uni2032"/><!-- PRIME -->
    </cmap_format_4>
  </cmap>

</ttFont>

```

What we see is that our `uE00D4` is nowhere to be found, and instead the glyph carries the never-asked-for `u2032` codepoint.

[unicode-test.zip](https://github.com/adobe-type-tools/afdko/files/5718440/unicode-test.zip)

punchcutter:
Fixes #1273 

The code was setting the single Unicode value taken from parsing the final name and then moving on to the next glyph. I swapped the order so it first checks if there are multiple Unicodes to be assigned. One other thing we could consider is the following scenario

`uniABCD    uniABCD    uni1234,uni5678`

Here we will assign U+1234 and U+5678, but not U+ABCD so we could maybe warn if the name looks like a Unicode value and we don't assign it.


punchcutter:
Tests seem to only fail on Linux which I think was mentioned somewhere else, right?

skef:
The Linux failures look like they're probably specific to this PR. Haven't determined why yet.

tests/makeotfexe_test.py
```diff
@@ -707,6 +707,19 @@ def test_negative_internal_leading_bug1227():
             b'win.ascent + win.descent < unitsPerEm') in output
 
 
+def test_goadb_multiple_unicodes_bug1273():
+    input_filename = "bug1273/font.pfa"
+    ttx_filename = "bug1273.ttx"
+    goadb_filename = 'bug1273/GLyphOrderAndAliasDB'
```
skef, line 713: The "L" in "GLyph" is capitalized here but not in the filename above. That's working on MacOS because its filesystem is case-insensitive but won't work in other contexts.

skef (approved):
This looks fine to me otherwise. Approving pending test code change.

punchcutter:
Aha, duh. I'll fix and push.

josh-hadley (approved):
Thank you for addressing this

PR #1633, opened 2023-04-06
[tx] revert missing cid fail to warn

kaydeearts:
## Description

In #1628 , we made some commits that would deprecate from FAIL to WARN when a glyph did not have a CID defined if it was a CID-keyed UFO. Zachary found some issues with this with some further testing. We'll do this in the near future, just wanted to make a release without this for now.  

## Checklist:


skef (approved):
lgtm

Issue #1636, opened 2023-04-12
[spot] name table LANG_TAG_REC_SIZE size definition.


NSGod:
In /afdko/c/spot/sfnt_includes/sfnt_name.h, you define `LANG_TAG_REC_SIZE` for `LangTagRecord`, but it takes the `SIZEOF()` from the `NameRecord` struct instead of the `LangTagRecord`.

```
typedef struct
{
    Card16 platformId;
    Card16 scriptId;
    Card16 languageId;
    Card16 nameId;
    Card16 length;
    Card16 offset;
} NameRecord;
#define NAME_REC_SIZE (SIZEOF(NameRecord, platformId) + \
                       SIZEOF(NameRecord, scriptId) +   \
                       SIZEOF(NameRecord, languageId) + \
                       SIZEOF(NameRecord, nameId) +     \
                       SIZEOF(NameRecord, length) +     \
                       SIZEOF(NameRecord, offset))

typedef struct
{
    Card16 length;
    Card16 offset;
} LangTagRecord;
#define LANG_TAG_REC_SIZE (SIZEOF(NameRecord, length) + \
                           SIZEOF(NameRecord, offset))
                                      ^_ should be LangTagRecord
```

Note that this bug doesn't actually have any effect because the `length` and `offset` fields of those structs both happen to have the same size.

PR #1640, opened 2023-04-13
[spot] fix name table LANG_TAG_REC_SIZE size definition.

NSGod:
Correcting this definition doesn't actually change the value, since the sizes of the `length` and `offset` fields of `LangTagRecord` are identical to those of the `NameRecord`.

## Description

This is a bug fix for #1636 name table LANG_TAG_REC_SIZE size definition.

## Checklist:

Since this doesn't actually have any effective change, I didn't run the tests (which I still haven't been able to get fully working yet).


miguelsousa (approved):
Thank you!

PR #1575, opened 2022-11-10
[WIP][tx][libxml2-improved] Move FDArray from fontinfo.plist to lib.plist

kaydeearts:
## Description

Does what #1380 intended, but rewritten for the new libxml2 improvements + other considerations.
Opening a PR early to have as a draft so we can discuss & revise as I work on this! 🚀 

Note: I branched off of #1568 for now, will fix this one #1568 merges into develop (this is why you're seeing extra commits)

## To-Dos:
- [x] ufowrite: write FDArray in lib.plist instead of fontinfo.plist
- [ ] uforead: parse FDArray from lib.plist as well as fontinfo.plist(?)
- [ ] .glif files: current solution is that iFD & CID is held in glif files. Is this what we want?
- [ ] update tests to reflect the changes

## Checklist:

- [ ] I have followed the [Contribution Guidelines](https://github.com/adobe-type-tools/afdko/blob/develop/CONTRIBUTING.md)
- [ ] I have added **test code and data** to prove that my code functions correctly
- [ ] I have verified that new and existing tests pass locally with my changes
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

PR #1642, opened 2023-04-19
[uforead.c] Revert glifRec array sort

### LINKED ISSUE #1641: [makeotf] wrong codepoint assignment
I just installed afdko 3.9.4 in the context of a workshop, and `makeotf` seems to write arbitrary code points for a very simple exercise (attached). The project is built without GlyphOrderAndAliasDB, the resulting code point assignment looks like this:

Screenshot 2023-04-17 at 14 20 32
Screenshot 2023-04-17 at 14 20 41

[font.ufo.zip](https://github.com/adobe-type-tools/afdko/files/11250622/font.ufo.zip)

`makeotf -r` for Source Serif Regular looks like this:
Screenshot 2023-04-17 at 14 17 32

kaydeearts:
## Description

To fix #1641.

In #1575 in findGLIFRecByName function, I switched from using a for loop to using bsearch when finding a specific glifRec. 
I had also added a qsort function since bsearch requires the array to be sorted, but I sorted by glyphName, which messed up the glyphOrder. We already were sorting a different way but sorting by glyphOrder instead.

## Checklist:


c/shared/source/uforead/uforead.c
```diff
@@ -1228,7 +1228,16 @@ static GLIF_Rec* findGLIFRecByName(ufoCtx h, char *glyphName)
 
     GLIF_Rec* glif = bsearch(glifKey, h->data.glifRecs.array, h->data.glifRecs.cnt, sizeof(GLIF_Rec), glifRecNameComparator);
```
skef, line 1229: I'm not sure if you're keeping this code around for future use but if we're reverting we should probably comment the bsearch and related lines out. (And I suppose if we do keep the lines we should probably keep the removed qsort in a comment as well.)

kaydeearts: Ah sorry about this — I pushed this in a rush earlier as I was leaving for something & was looking over my PR again just now (should've get it in draft mode until I had re-reviewed). Needs some cleaning, thanks!

skef (approved):
LGTM

kaydeearts:
@skef Ah I added one more test right after your review — if that test looks good, I'll go ahead & merge!

skef (approved):
Private secret just between us: I don't usually review test code unless I'm worried about uncovered cases, which is rare. If the code for a test is lower quality it will either pass when it shouldn't (unlikely) or someone will bring it to our attention when it fails.

PR #1644, opened 2023-04-19
[hotconv] 'name': fix memory leak in addName()

### LINKED ISSUE #1643: [hotconv] 'name': memory leak in addName()
I think I may have found a leak in `name.c -> addName()`

In name.c, line 168:

```
static void addName(nameCtx h,
                    unsigned short platformId,
                    unsigned short platspecId,
                    unsigned short languageId,
                    unsigned short nameId,
                    size_t length, char *str) {
    char *dst;
    char *newStr;
    newStr = MEM_NEW(h->g, length * sizeof(char)); <<-- ALLOCATED
    NameRecord *rec;
    int index;
    int omitMacNames = (h->g->convertFlags & HOT_OMIT_MAC_NAMES); /* omit all Mac platform names. */

    if (platformId == HOT_NAME_MAC_PLATFORM) {
        if (omitMacNames)
            return;
    }

    if (omitMacNames && (platformId == HOT_NAME_MAC_PLATFORM)) {
        return;
    }

```
First, it looks like the last 3 lines — and the 4 lines before that — effectively do the same thing: return early if it's Mac and we're not writing Mac platform names. However, `newStr` is allocated before that `return` and not released until near the end of the function, which would cause a leak, I believe.

NSGod:
## Description
This fixes #1643 [hotconv] 'name': memory leak in addName().

## Checklist:

As I've mentioned before, I still don't have testing figured out locally on my end, so I'll submit this and see how the tests do on your end.

### DIFF
--- a/c/makeotf/lib/hotconv/name.c
+++ b/c/makeotf/lib/hotconv/name.c
@@ -173,19 +173,15 @@ static void addName(nameCtx h,
                     size_t length, char *str) {
     char *dst;
     char *newStr;
-    newStr = MEM_NEW(h->g, length * sizeof(char));
     NameRecord *rec;
     int index;
     int omitMacNames = (h->g->convertFlags & HOT_OMIT_MAC_NAMES); /* omit all Mac platform names. */
 
-    if (platformId == HOT_NAME_MAC_PLATFORM) {
-        if (omitMacNames)
-            return;
-    }
-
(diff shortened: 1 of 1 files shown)


NSGod:
Closing to resubmit a pull request w/ updated (rebased on develop) fix.

Issue #1647, opened 2023-05-03
[hotconvert] BUG?/clarification on how HOT_ADD_STUB_DSIG is supposed to work.


NSGod:
I've encountered what I believe might be a bug, though first I'd thought I'd get clarification on how the `HOT_ADD_STUB_DSIG` flag is supposed to be used. I don't know enough about how `makeotfexe` leverages `hotconvert` to know if it's possible that it would ever encounter this problem or not. 

I have an app that's leveraging the hotconvert library as a framework to do the conversion of fonts to OTF data. I'm creating a single `hotCtx` and using it to convert more than one font. I'd like for each converted font to have a stub `'DSIG'` table. It appears the convert flags are reset after 

The following is an excerpt from a sample project I set up at [NSGod/afdko/md-dsig-stub-bug-demo](https://github.com/NSGod/afdko/tree/md-dsig-stub-bug-demo) to demonstrate the problem I'm encountering (apologies but it's written in Objective-C):

```
static hotCtx                _hotCtx;
static hotCallbacks     _hotCallbacks;
static mdHotCallbacks { ....};

int main(int argc, const char * argv[]) {
   @autoreleasepool {
   srcDataHandle = [[FFMutableDataHandle alloc] init];
   CFFDataHandle = [[FFMutableDataHandle alloc] init];
   OTFDataHandle = [[FFMutableDataHandle alloc] init];
   
   _hotCallbacks = mdHotCallbacks;
   _hotCallbacks.ctx = NULL;
   _hotCtx = hotNew(&_hotCallbacks);
   hotSetConvertFlags(_hotCtx, HOT_ADD_STUB_DSIG);
   
   NSURL *tempDirURL = [[NSURL fileURLWithPath:NSTemporaryDirectory()]
                                                  URLByAppendingPathComponent:@"md-dsig-stub-fix-078j"];
   NSURL *pfaURL = [tempDirURL URLByAppendingPathComponent:@"type1.pfa"];
   
   for (NSUInteger i = 0; i < 4; i++) {
      srcDataHandle.data = [NSData dataWithContentsOfURL:pfaURL];
      
      int psInfo = 0;
      hotReadFontOverrides fontOverrides = {0};
      
      char *fontName = hotReadFont(_hotCtx, HOT_SUPRESS_HINT_WARNINGS |
                                                            HOT_NO_OLD_OPS, &psInfo, &fontOverrides);
      (void)fontName;
      
[...]

PR #1648, opened 2023-05-03
[hotconv] reset dsigCnt in hotReuse() so subsequent conversions add full stub 'DSIG' table.

NSGod:
## Description
This is a patch to fix bug #1647. Basically, I moved the static `dsigCount` variable outside the `refillDSIG()` function, and reset it in `hotReuse()`.

## Checklist:

I think I finally have all local tests working except for `test_infinite_loop_with_dflt_lookups_bug965()`. That test fails before and after my changes, so I don't think it'd be related.

Unlike the previous bugs I've encountered , this strikes me as one that could probably use a test? The thing is, as I've mentioned elsewhere, I'm not sure if it's possible to trigger this bug using existing tools, and I'm still in the process of learning Python enough to write a test. If you think it warrants a test and could create one, that would certainly help in my learning process.

### DIFF
--- a/c/makeotf/lib/hotconv/hot.c
+++ b/c/makeotf/lib/hotconv/hot.c
@@ -851,6 +851,8 @@ static void setVBounds(hotCtx g) {
     }
 }
 
+static unsigned int dsigCnt = 0;
+
 static void hotReuse(hotCtx g) {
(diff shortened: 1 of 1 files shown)


kaydeearts (approved):
Thank you!

PR #1651, opened 2023-05-09
[spec] Multiple substitution with classes syntax

khaledhosny:
## Description

Document syntax for using glyph classes in multiple substitution rules. The syntax used by [GlyphsApp][1], and recently also supported by [feaLib][2].

## Checklist:

[1]: http://handbook.glyphsapp.com/en/layout/multiple-substitution-with-classes/
[2]: https://github.com/fonttools/fonttools/pull/3103

### DIFF
--- a/docs/OpenTypeFeatureFileSpecification.md
+++ b/docs/OpenTypeFeatureFileSpecification.md
@@ -1602,18 +1602,37 @@ Omitting the `by` clause is equivalent to adding `by NULL`.
 <a name="5.b"></a>
 ### 5.b. [GSUB LookupType 2] Multiple substitution
 
-A Multiple Sub rule is specified as:
+A Multiple Sub rule is specified in one of the following formats:
 
 ```fea
-substitute <glyph> by <glyph sequence>;
+substitute <glyph> by <glyph sequence>;            # Format A
+substitute <glyphclass> by <glyph sequence>;       # Format B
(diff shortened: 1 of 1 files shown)


khaledhosny:
> @khaledhosny Actions and other GH systems [are experiencing an outage](https://www.githubstatus.com/); that's why CI hasn't run yet.

This is only documentation change, though, so probably there is not much for the CI to catch anyway.

miguelsousa:
I see. Do you mind creating an issue to request supporting the new formats in makeotf?

khaledhosny:
https://github.com/adobe-type-tools/afdko/issues/1652

skef:
This looks like a good extension to the syntax and we will likely add it. For now the addition seems lower priority than the ongoing VF-related work, so it will happen later. We generally try to keep the FDK current with the grammar so the doc update will happen at the same time #1652 is resolved.

skef:
@khaledhosny  I'm going to start directing this sort of proposal to https://github.com/adobe-type-tools/feature_file_workshops for (eventual) discussion by the broader font tools community.

How would you feel about making a PR there with this proposal? You have the language worked out already. Ideally you'd take a stab at the grammar and provide some examples, but that isn't mandatory. 

Or, stepping up a level, could you take that possibility up with whoever else is relevant on your side of the fence and let us know what you're thinking, either via an issue, PR, or using the address at the link above?

Issue #1653, opened 2023-05-14
[hotconv] negative TypoLineGap warning checks TypoAscender


NSGod:
In `c/makeotf/lib/hotconv/hot.c` is the following code starting at line 760:

```
    if (!OVERRIDE(font->TypoLineGap)) {
        font->TypoLineGap = IS_CID(g)
                                ? font->TypoAscender - font->TypoDescender
                                : EM_SCALE(1200) - font->TypoAscender + font->TypoDescender;
    }

    /* warn if the line gap is negative. */
    if (font->TypoAscender < 0) {
        hotMsg(g, hotWARNING, "The feature file OS/2 override TypoLineGap value is negative!");
    }
```

It looks like it's supposed to be checking the line gap and warning if it's negative, but it's actually checking `font->TypoAscender < 0` not `font->TypoLineGap < 0`.

PR #1654, opened 2023-05-14
[hotconv] fix warning for negative TypoLineGap.

NSGod:
## Description
This is a fix for #1653.

## Checklist:

I ran tests locally and they all passed except the infinite loop one which always crashes for me.

It seems like it might be useful to have a new test that checks for that warning but I don't yet know enough about the testing framework to be able to write one myself.


kaydeearts (approved):
Thank you!

PR #1668, opened 2023-06-30
[buildcff2vf] Don't return zero when input requires compatibilization

### LINKED ISSUE #1667: [buildcff2vf] returns 0 when no vf is built
`buildcff2vf` can identify when glyphs are not 100% compatible. In that case, I get a message  `The input set requires compatibilization. Please try again with the -c (--check-compat) option.` – and the program ends – with return value 0.

Is it wrong to expect a non-zero return value here?

josh-hadley:
## Description

What the title says. Fixes #1667.

## Checklist:

- [x] I have ~added~ updated **test code and data** to prove that my code functions correctly

### DIFF
--- a/python/afdko/buildcff2vf.py
+++ b/python/afdko/buildcff2vf.py
@@ -23,7 +23,7 @@ from fontTools.varLib.cff import (CFF2CharStringMergePen,
 
 from afdko.fdkutils import validate_path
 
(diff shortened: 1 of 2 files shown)


miguelsousa (approved):

frankrolf:
Thank you!

Issue #1662, opened 2023-06-13
[makeotf] PUA unicodes automatically assigned to /dieresisacute and /dieresisgrave


BoldMonday:
When compiling source files with makeotf, glyphs named **dieresisacute** and **dieresisgrave** will automatically get PUA unicode values of **0xF6D7** and **0xF6D8**.

The supplied GlyphOrderAndAliasDB file does not specify any unicode value for these glyphs and that's also what we expect to be in the generated fonts:
```
[...]
dieresisacute	dieresisacute
[...]
dieresisgrave	dieresisgrave
[...]
```
Is there a way to disable any of this automatic mapping by makeotf?

frankrolf:
`makeotf` follows the AGL for assigning code points, potentially including this entry:
https://github.com/adobe-type-tools/agl-aglfn/blob/master/glyphlist.txt#L155-L156

Two ways to get around this:
- change the final name (left column)
- assign a codepoint override (3rd column)

frankrolf:
One example:
https://github.com/adobe-fonts/source-serif/blob/main/Roman/GlyphOrderAndAliasDB#L815-L818

BoldMonday:
Thanks for the example.

I wish there was a way to disable any of this automatic mapping when a GlyphOrderAndAliasDB file is supplied during compilation.
At least it would be helpful if makeotf would report automatic mapping during compilation. Now we suddenly ended up with a font containing PUA unicodes where we had no idea where they came from. And subsequently a faulty font because the OS/2 unicode ranges did not have a flag for the PUA range.

frankrolf:
I think this might be a “careful what you wish for” kind of situation – the AGD mapping is also the reason for glyph names like `Aacute` to receive the correct code point.

Of course, it’s odd that the AGD would map to PUA, not sure what the reason for that originally was.

I made a little script for you to verify which glyphs go through the AGD:
```
from afdko import agd, fdkutils
from pathlib import Path
f = CurrentFont()
agd_txt = Path(fdkutils.get_resources_dir()) / 'AGD.txt'

with open(agd_txt, "r") as agd_blob:
    agd_data = agd_blob.read()

agd_dict = agd.dictionary(agd_data)  # this is called dictionary but doesn’t seem to be a proper dict object :-/

for gname in f.templateGlyphOrder:
    agd_glyph = agd_dict.glyphs.get(gname)
    if agd_glyph:
        print(agd_glyph.uni, gname)
    else:
[...]

PR #1712, opened 2023-09-29
remove PUA code points from resource files

frankrolf:
Upon a rebuild of the FDK, this should probably address #1662.

## Description

This commit is deleting any glyph name associated with a PUA code points from AGD.txt, and (probably more importantly) agl2uv.h.

The assumption is that `makeotfexe` will no longer unexpectedly assign PUA code points.
I don’t have a way to test this locally (or I don’t know) – would be curious to figure out what the next steps are.


frankrolf:
I guess this means that some of the expected CJK fonts contain PUA glyphs :-)

josh-hadley:
> I guess this means that some of the expected CJK fonts contain PUA glyphs :-)

No — these test failures are caused by a change in fontTools v4.43.0, unrelated to your changes around removing PUAs.

josh-hadley:
@frankrolf once #1715 is merged, rebase this and it should pass. I'm still not convinced of the need to change AGD.txt but I guess it makes sense to keep its contents in sync with what `makeotf[exe]` is doing.

josh-hadley (approved):

PR #1716, opened 2023-09-30
add guards for h->otl == NULL before calling otlSubtableAdd

josh-hadley:
## Description

Add guards for h->otl == NULL in GSUB and GPOS code. Previously this was not explicitly checked, but could be NULL resulting in segfault.

## Checklist:

- [ ] ~~I have added **test code and data** to prove that my code functions correctly~~


skef (approved):
LGTM

PR #1749, opened 2024-07-08
Correct inequality in measuring adjacent BlueValue zones.

skef:
See recent Slack conversation in Rally channel.


josh-hadley (approved):
👍

PR #1751, opened 2024-07-11
Fix a couple otfautohint bugs found by testing with Momochidori

skef:
This fixes two bugs found by hinting Momochidori. One fix removes a misguided rounding, the other fixes the use of an index variable.

This also adds a new "loose-overlap-mapping" to make the reverse mapping of path-elements-with-overlap-removed back onto original-path-elements less strict. This helps with path elements that are big "loops", which booleanOperations doesn't always preserve the shape of as well. 

There are some other things we could try to avoid these kinds of heuristics, including switching from booleanOperations to Skia PathOps for overlap removal. However, the autohinter is just sort of OK, so I'm wary of putting hours into part A for superficial reasons just to avoid "formal" errors that may not have much of an impact on output. This flag was easy to add and mostly works around the problem. (You'll still get some missed mappings with the flag, just fewer of them. They're mostly harmless -- a bunch of things have to stack up for a given unmapped segment to make a difference.)

## Checklist:

### DIFF
--- a/python/afdko/otfautohint/__main__.py
+++ b/python/afdko/otfautohint/__main__.py
@@ -343,6 +343,7 @@ class HintOptions(ACOptions):
         self.writeToDefaultLayer = pargs.write_to_default_layer
         self.maxSegments = pargs.max_segments
         self.verbose = pargs.verbose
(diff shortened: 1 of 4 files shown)


josh-hadley (approved):
👍

PR #1754, opened 2024-08-30
Update makeinstancesufo to use latest ufoProcessor

punchcutter:
This makes use of the "new" way to use ufoProcessor with UFOOperator. This allows passing the `--strict` flag from `makeinstancesufo` to `ufoProcessor` which passes to `fontMath`.  From `fontMath.mathGlyph`:

```
strict (bool): when set to False, offcurve points will be added to all
                straight segments to improve compatibility. Any offcurves that are
                still on-point will be filtered when extracted. When set to True,
                no offcurves will be added or filtered.
```

This means when making an "Instance" to be inserted back into the designspace we won't run into the situation where points are filtered out, making the new instance incompatible with the original sources.


punchcutter:
Oh yeah, tests fail because we need an update to ufoProcessor. Either that or change makeinstancesufo.

punchcutter:
This introduces a number of changes to the expected output UFOs, but I believe all of it makes sense and in some cases is now "more correct" so that's a good thing.

skef (approved):
This looks good to me.

skef:
@josh-hadley I believe you were more involved in the discussions about this -- everything look good to you?

josh-hadley:
> @josh-hadley I believe you were more involved in the discussions about this -- everything look good to you?

Right, I've been informally chatting with @punchcutter about this. This looks good to me and will solve some longstanding annoyances.

Issue #1761, opened 2024-11-15
[makeoffexe] 4.0.2 fails on Linux


takaakifuji:
MakeOTF causes the following error on the GitHub Actions runner, which runs Ubuntu 20.04.6 LTS and CPython 3.12.7:

```
$ git clone 'https://github.com/adobe-fonts/adobe-blank.git' && cd adobe-blank && source COMMANDS.txt
```

```
makeotf [Warning] Did not recognize the ordering 'Identity'. The Mac cmap script id will be set to Roman (0).
makeotf [Warning] the feature file does not contain a 'vert' feature features.
makeotf [Warning] Could not find an Adobe CMAP Mac encoding mapping file. If you want a Mac cmap subtable, please specify the Mac CMAP encoding file with the '-cm' option.
makeotf [Warning] Could not find a Unicode Variation Sequence file in the expected '/home/ubuntu/.pyenv/versions/orcatools-3.12.7/lib/python3.12/site-packages/afdko/resources' subdirectory.
makeotf [Note] setting the USE_TYPO_METRICS OS/2 fsSelection bit 7 from fontinfo keyword.
makeotf [Note] setting the WEIGHT_WIDTH_SLOPE_ONLY OS/2 fsSelection bit 8 from fontinfo keyword.
makeotf [Note] setting the OBLIQUE OS/2 fsSelection bit 9 from fontinfo keyword.
makeotfexe processing font <AdobeBlank>
WARNING: cmap{plat=3,script=1,lang=0}: cmap format 4 subtable was truncated to 2 segments, due to overflow with non-truncated version
WARNING: cmap{plat=1,script=0,lang=0}: empty encoding
WARNING: Macintosh Adobe CMap not seen
WARNING: Unicode variation Selector File not seen
terminate called after throwing an instance of 'std::system_error'
  what():  Unknown error -1
Error executing command 'makeotfexe -f /tmp/tmpfb7gpzv7 -o /home/ubuntu/adobe-blank/AdobeBlank.otf -omitMacNames -ff /home/ubuntu/adobe-blank/features -mf /home/ubuntu/adobe-blank/FontMenuNameDB -r -stubCmap4 -ch /home/ubuntu/adobe-blank/UnicodeAll-UTF32-H -lic ADOBE'
makeotf [Error] Failed to build output font file '/home/ubuntu/adobe-blank/AdobeBlank.otf'.
sfntedit [FATAL]: file error <No such file or directory> [AdobeBlank.otf]
[...]

PR #1763, opened 2024-11-15
Try to work around pthread related static linking problem (#1761)

skef:


skef:
I'm dealing with this as part of a bulk edit.

PR #1772, opened 2025-01-24
More otfautohint rounding fixes

skef:
Complete the rounding fixes in otfautohint (problems found with Momochidori testing).


skef:
Something is causing a difference between the codeQL environment and the AFDKO Test Suite environment, such that one set of test files works in one and not the other. I haven't been able to figure out what's making the difference yet.

PR #1774, opened 2025-04-10
Fix handling of CID-keyed UFOs with no FDSelect

punchcutter:
Originally we based the CID-keyed UFO workflow on all of the Adobe CJK fonts which have multiple font dicts, but we failed to account for the possibility of a totally valid CID-keyed file with no font dicts (no FDSelect because there are no FDArrays to point to). This adjusts the reading and writing of UFOs to properly handle this. The documentation is also updated so that we do not absolutely require the postscriptFDArray key in the lib.plist since it's actually optional.


skef (approved):
LGTM

Issue #1632, opened 2023-03-30
[tx] Extract t1 from UFO fails but extracting cff works


NightFurySL2001:
When passing the following UFO fonts into `tx -t1`, the program reports an error as below:
```
tx.exe: --- .\orig-pearl-tangyuan.ufo
tx.exe: (ufr) parsing failed
```
One is directly exported from FontLab, another (-2) is slightly modified. Both UFO are able to work with `tx -cff` and then `tx -t1`, but the glyph ordering is lost in the process.

Both fonts are able to be built with Google's `fontmake` so there shouldn't be any problem with the UFOs.

[orig-pearl-tangyuan.zip](https://github.com/adobe-type-tools/afdko/files/11112054/orig-pearl-tangyuan.zip)
[orig-pearl-tangyuan-2.zip](https://github.com/adobe-type-tools/afdko/files/11112055/orig-pearl-tangyuan-2.zip)

punchcutter:
There's this contour way on the right side of uni8D9F that's causing it to fail. When converting to t1 it doesn't know what to do with that. 
<img width="1192" alt="uni8D9F" src="https://user-images.githubusercontent.com/11257204/229291264-33129207-7b23-4a01-951a-dd1d0dec3dd9.png">

Delete it and everything is fine. We should give a better error message since "parsing failed" doesn't really tell us anything .

punchcutter:
If you converted with `-cff` first you should have seen the warning `tx: (cfw) no initial moveto (inserted) <uni8D9F>`

frankrolf:
Ran into this today with a UFO that contained a single quadratic segment (among 3725 cubic segments).

The error message was `Command '['tx', '-t1', 'font.ufo', '/var/folders/7z/vl3dc3kd07bfz1q_5ybyk5wm0000gn/T/tmphui0y0jo']' died with <Signals.SIGSEGV: 11>.`, which is cryptic.

kaydeearts:
Thanks for this submission! The issue Frank commented above was similar but exercised a different part of the code (and a fix for it was just merged into develop) but I've started investigating this specific issue as well

punchcutter:
The original issue was a contour with a single point in it. Here is a much smaller test case: [single-point-contour-failure.ufo.zip](https://github.com/adobe-type-tools/afdko/files/11417782/single-point-contour-failure.ufo.zip)

[...]

Issue #1540, opened 2022-07-15
[tx] UFO glif parsing fails on contours with a single point


punchcutter:
Parsing shouldn't just fail here. In this case the single points shouldn't be there, but tx could issue a warning and continue instead of failing with `tx: (ufr) parsing failed` and no other information.

kaydeearts:
@hatchzo also added:
> The zigzag/loop paths should be corrected by design and tx should report detailed information on parse errors.
But I think the most serious problem is checkoutlineufo. It seems that those single point contours are generated while running checkoutlineufo.
I made a test ufo file that only collect glyphs that cause errors and then run that :
`checkoutlinesufo -e  -w -o test-check.ufo test.ufo`

skef:
@punchcutter it looks like #1786 does address #1632 but this issue also asks to change the error to a warning. Is that no longer desired/needed?

punchcutter:
I don't think we need anything else. If we get this new better message telling us which glyph had a problem then the user can check that glyph and try again.

PR #1786, opened 2025-06-04
Added some basic tx -t1 error messaging for #1540 and #1632

punchcutter:
Added some minimal error messaging to handle the errors in #1540 and #1632. This gives us the problem glyph instead of the generic `parsing failed` which tells us nothing about why or where. 

```
tx: --- orig-pearl-tangyuan.ufo
  ERROR: Error in glyphEnd for glyph 'uni8D9F': inappropriate function call or arguments
  FATAL: parsing failed
```


punchcutter:
Just realized this only happens in these test files because the stray point is the first contour. If it's not the first then it doesn't fail. Ideally we could check in more detail to report a better message, but I still think this is at least a decent start to report something rather than nothing.

PR #1793, opened 2025-06-16
docs updates [skip ci]

frankrolf:
minor updates to AFDKO-Overview.md

### DIFF
--- a/docs/AFDKO-Overview.md
+++ b/docs/AFDKO-Overview.md
@@ -37,7 +37,8 @@ This script will use a .designspace file and the output OTFs of `buildmasterotfs
 
 ### `makeinstancesufo`
 
-This script will generate UFO instances from a .designspace file, and a set of compatible UFO masters. More info on designspace files and interpolation systems can be found at [Superpolator.com](http://superpolator.com/) and [DesignSpaceDocument Specification](https://github.com/fonttools/fonttools/tree/master/Doc/source/designspaceLib).
+This script will build UFO instances from a .designspace file and a set of compatible UFO masters, to prepare for building static OTF fonts. By default, the instances are run through `checkoutlinesufo` (for overlap removal) and `otfautohint` (for PS hinting). Finally, the UFO instances are normalized.  
+More info on designspace files and interpolation systems can be found at [Superpolator.com](https://superpolator.com/designspace.html) and [DesignSpaceDocument Specification](https://fonttools.readthedocs.io/en/latest/designspaceLib/index.html).
 
 ### `mergefonts`
 
--- a/docs/AFDKO-Overview.md
+++ b/docs/AFDKO-Overview.md
@@ -137,7 +137,7 @@ The auto-hinting program will report at length about hinting issues. Some of the
 
 ### `checkoutlinesufo`
 
-This tool will check the quality of the glyph outline data, and should always be used. It is very good at detecting serious problems, such as overlap and incorrect path direction. It is overly enthusiastic about finding a number of smaller issues, but is right often enough to be worth checking all the error messages. It can also fix the problems it finds, but you should always check any glyphs that it changed - the fixes are not always better than the original problem.
+This tool can be used to check the general quality of glyph outline data. The `-e` option can be passed to remove overlaps. When working with UFO files, the results of overlap removal are written to a new layer (`com.adobe.type.processedglyphs`), the original glyph is not touched.
 
 ### `comparefamily`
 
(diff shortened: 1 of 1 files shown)


docs/MakeOTFUserGuide.md
```diff
@@ -38,90 +41,81 @@ MyFontFamily/
...
+Subroutinization is a process by which common elements in a set of glyphs are decomposed in separate subroutines. This can reduce the size of the final font but can require extreme amounts of memory and time for a large font, such as CID fonts. `makeotf` may need as much as 64 Mbytes of memory for large Roman fonts, and will do most with only 32 Mbytes, but it may need 768 Mbytes of memory to subroutinize a 5-Mbyte CID font. Subroutinizing is reasonably fast when done all in memory: a few minutes for a Roman font, half an hour to three hours for a CID font. However, if the system must use virtual memory, the time required can increase by a factor of 20 or more. Subroutinizing is likely to be useful for Roman fonts, and for Korean CID fonts. Japanese and Chinese CID fonts usually only yield a few percent size reduction with subroutinizing, due to fewer repeating path elements.
```
skef, line 112: This can be:

> Subroutinization is a process by which common elements in a set of glyphs are decomposed in separate subroutines, reducing the size of the final font. Subroutinizing should take at most a few minutes for a typical Roman font, but could take up to 3 hours for a large CJK font.
Subroutinizing is likely to be most useful for Roman and Korean fonts. Japanese and Chinese fonts usually only yield a few percent size reduction with subroutinizing, due to their being fewer repeating path elements.

docs/MakeOTFUserGuide.md
```diff
@@ -239,30 +233,25 @@ Versions of MakeOTF prior to FDK 2.5 used a similar syntax in the FontMenuNameDB
...
-Fields within a line are tab separated
-(technically any amount of a whitespace works but a single ASCII TAB is preferred).
-Blank lines are ignored.
-Lines beginning with `#` are comments and are also ignored.
-The first field is the final glyph name to be used in the output font.
-The second field is the ‘friendly’ name used in the source font data. 
-The third field is a Unicode value, specified in the form `uniXXXX` or `uXXXX[XX]` (see [note](#unicode_note)). 
-One may specify more than one Unicode value for a glyph by giving a comma separated list of values, for example: `uni0020,uni00A0`. 
+The GOADB file is used to rename and to establish an order for the glyphs in a font. It is a simple text file with one line per glyph name. Each line contains at least two columns, and optionally a third column.  
+Colums are tab-separated (technically any amount of a whitespace works but a single tab character is preferred).
```
skef, line 241: Columns

skef, line 241: And "recommended" rather than "preferred".

docs/MakeOTFUserGuide.md
```diff
@@ -386,38 +373,10 @@ When bit 9 is on, programs that support the CSS2 font specification standards wi
...
-The glyphs are synthesized from the MM fonts which MakeOTF has built-in. It will try to match the glyph width of the zero and the dominant stem width of the target font. However, MakeOTF cannot stretch the MM font data to match very thick strokes, very wide glyphs, and it cannot match the design’s stem contrast.
-
 ---
 
 #### Document Version History
 
 Version 1.0 - Initial version  
-Version 1.5 - Revisions for MakeOTF v2.5 - 2016  
+Version 1.5 - Revisions for `makeotf` v2.5 - 2016  
 Version 2.0 - Convert to Markdown, many minor updates and fixes - Josh Hadley, October 2019
```
skef, line 382: Seems like this is enough change to at least bump to 2.1, no?

docs/MakeOTFUserGuide.md
frankrolf: I have a feeling that talking about lengths of time won’t age well. I’ve never seen “minutes” needed for Western fonts – usually, the operation is much quicker.

docs/MakeOTFUserGuide.md
frankrolf: I just checked this – indeed, any whitespace separation works. Not sure why we’d recommend one over the other then?

skef: I'm fine with getting rid of that too. "Preferred" didn't make sense -- what would be doing the preferring? If we want to recommend, we can recommend. Or we can delete that.

docs/MakeOTFUserGuide.md
frankrolf: indeed :-)

PR #1799, opened 2025-08-27
don’t refer to non-existent user guide PDF

frankrolf:
makeotf help update.

I was looking for documentation on the fontinfo file, when I noticed this little oversight.


skef:
I'm fixing this as part of a "bulk" update.

PR #1803, opened 2025-09-16
[rotatefont] support writing UFOs

### LINKED ISSUE #1802: [rotatefont] cannot write UFO files
```plain
# abc.map
.notdef	.notdef	None	0	0
A	A	None	0	0
B	B	None	0	0
C	C	None	0	0
```

```
$ rotatefont -ufo -rtf abc.map SourceSans3-Regular.otf abc.ufo
$ tx -ufo SourceSans3-Regular.otf SourceSans3-Regular.ufo && rotatefont -ufo -rtf abc.map SourceSans3-Regular.ufo abc.ufo
```

This should work and produce a UFO with 4 glyphs according to the mapping file but it doesn't. The former causes a segmentation fault and the latter produces a non-subset UFO.

Not sure if rotateFont is originally intended to support formats other than t1/cff, but it's nice to have and it looks like the codebase is designed to do so.

takaakifuji:
## Description

Intended to fix #1802. Only rotateFont.c has been modified and it should not affect any other part of AFDKO.

## Checklist:

### DIFF
--- a/c/rotatefont/source/rotateFont.c
+++ b/c/rotatefont/source/rotateFont.c
@@ -43,6 +43,7 @@ typedef struct
     float origMatrix[6];            /* rotation/translation matrix specified by user. */
     float curMatrix[6];             /* rotation/translation matrix currently in use. */
     abfGlyphCallbacks savedGlyphCB; /* originally chosen writer call-backs */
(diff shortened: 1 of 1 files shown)


skef:
This fix is going in through the addfeatures branch. Closing this PR.

PR #1833, opened 2026-04-23
Correct keywords for table/feature blocks

signalwerk:
The following keywords were missing from the list (but specified in the corresponding table/feature):

* `nameid`
* `FSType`
* `name`
* `WeightClass`
* `WidthClass`
* `LowerOpSize`
* `UpperOpSize`
* `FamilyClass`

The sorting of the keywords in the list was also not consistent (not alphabetical by keyword, not alphabetical by table/feature name, almost order of appearance in the specification). It is now sorted by the order of appearance in the specification (8.b, 9.a, 9.b, …).

### DIFF
--- a/docs/OpenTypeFeatureFileSpecification.md
+++ b/docs/OpenTypeFeatureFileSpecification.md
@@ -223,49 +223,43 @@ table/feature blocks:
 
 | keyword | table | implemented |
 | -- | -- | -- |
-| [`HorizAxis.BaseScriptList`](#9.a) | BASE table | ✅ |
+| [`sizemenuname`](#8.b) | size feature | ✅ |
 | [`HorizAxis.BaseTagList`](#9.a) | BASE table | ✅ |
+| [`HorizAxis.BaseScriptList`](#9.a) | BASE table | ✅ |
 | [`HorizAxis.MinMax`](#9.a) | BASE table | ❌ |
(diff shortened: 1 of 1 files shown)


frankrolf:
Thank you for the update! Seems reasonable :-)

I’m curious – have you tested if these keywords have the intended effect when compiling a font?
```
    LowerOpSize <number>;
    UpperOpSize <number>;
    FamilyClass <number>;
```

Thanks again!

signalwerk:
@frankrolf Thank you for your reply. I was not testing this. I was working on a `.fea` syntax highlighter and I just realized the inconsistency in the documentation. Because it was not otherwise mentioned and the documentation implies it is working, I mentioned it as `implemented`.

frankrolf:
@signalwerk I hope you are aware of this: https://gitlab.com/saalen/highlight

I recently contributed .fea support to this package.

signalwerk:
@frankrolf thank you for pointing that out. I use support in [highlightjs](https://highlightjs.org/) but it's good to have your implementation. Great!

skef:
Given the timing this should go through the https://github.com/adobe-type-tools/feature_file_workshops process. @signalwerk do you have any interest in doing a PR over there, or would you like us to? Either is fine -- this is less of a change than just clean-up and therefore cosmically on us.

signalwerk:
@skef I created https://github.com/adobe-type-tools/feature_file_workshops/pull/2

Is that working for you like that?

PR #1843, opened 2026-06-27
addfeatures/hmtx: avoid unsigned integer underflow

eclairevoyant:
## Description

Bugfix related to https://github.com/NixOS/nixpkgs/issues/535868, `afdko` tests didn't pass on MacOS.

Because `size_t` is unsigned, if `hmtx.advanceWidth.size()` is `1` or `0`, or if the `hmtx.advanceWidth[i] == width` condition is always satisfied, then `i` will underflow (which would pass the loop condition, leading to an attempted out-of-bounds array access, which is UB). This PR prevents that.

## Checklist:


skef (approved):
LGTM

PR #1844, opened 2026-06-29
Fix spurious assertion when a high ghost stem is a glyph's first stem

philiptaron:
## Description

In `calcInstanceStems()` the per-stem loop initializes `lo = None` and `hi = None`, then computes one or both locations per stem. For a 'high' ghost stem (a hint anchored only at the top edge) only `hi` is computed; `lo` is then unconditionally overwritten by `hi` on the next line, so its incoming value is dead.

The 'high' branch nonetheless asserted `lo is not None`. When a high ghost is the glyph's *first* stem -- exactly the case for small diagonal accents (in the Cantarell typeface, that's acute, grave, caron, tonos, the Greek polytonic uni1Fxx stack, plus parenleft/right, slash, asterisk, quotes, ...) -- `lo` is still None from its initialization, so the assertion fired and aborted hinting of the whole font.

I got here by exploring the build failure of my NixOS system after afdko 5.0.1 was introduced as the `afdko` nixpkgs-wide. Suddenly [cantarell-fonts](https://gitlab.gnome.org/GNOME/cantarell-fonts/-/work_items/78) failed to build (NixOS/nixpkgs#535887).

After a decent amount of learning about typefaces and hinting, assisted by your friend and mine, Mr. Claude, I discovered the solution: drop the dead `lo` operand from the assertion, because `hi` is the only value actually read.

Claude added `tests/otfautohint_data/input/vf_tests/vf_highghost_first_stem.hinted.otf`, a minimal CFF2 variable font whose first glyphs are grave/acute accents with high ghost first stems. It is picked up by the existing globbed `test_vfotf round-trip` test, which raised `AssertionError` before this fix.

## Checklist:

- [x] I have commented my code, particularly in hard-to-understand areas (none required)
- [x] I have made corresponding changes to the documentation (none required)


skef (approved):
LGTM

PR #1848, opened 2026-07-07
Fix CID Mac cmap: suppress empty stub, fix -cl off-by-one

### LINKED ISSUE #1842: [addfeatures] Macintosh cmap subtable: unconditional empty stub for CID fonts; `-cl` value is written without the spec-required +1
**Environment:** afdko 5.0.1 (code links pinned to develop @ ca7161c; behavior identical). Building CID-keyed OTFs with:

```
afdko addfeatures -f font.cff -o out.otf -ff features -ch unicode.cmap -cs 25 -cl 33 -omitMacNames
```

(no `-cm`). The output contains a format 6 cmap subtable with 0 entries, **plat=1, script=25, lang=33**. Omitting `-cs` and `-cl` does not prevent it — the subtable is still emitted, as **plat=1, script=0, lang=0**.

### 1. The OT spec discourages platform-1 cmap subtables, but there is no way to turn it off

The spec (["Macintosh platform"](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#macintosh-platform-platform-id--1)) says:

> "Older Macintosh versions required fonts to have a 'cmap' subtable for platform ID 1. For current Apple platforms, use of platform ID 1 is **discouraged**."

Yet for CID fonts, `mapFill()` calls `makeCIDMaccmap()` unconditionally — the comment itself documents the stub behavior ([map.cpp#L2530](https://github.com/adobe-type-tools/afdko/blob/ca7161c/c/addfeatures/hotconv/map.cpp#L2530)):

> `makeCIDMaccmap(g); /* will make a Mac cmap with only a not def, if Macintosh Adobe CMap is not provided.*/`

Omitting `-cm` does not prevent the subtable from being created — it is still emitted, merely as an empty stub that maps nothing — and no option exists to turn it off. As for diagnostics: every CID build without `-cm` logs `"WARNING: Macintosh Adobe CMap not seen"` (a warning about the missing resource for a subtable the user never asked for), while omitting `-cs` or `-cl` produces no warning at all — the script and language fields simply default to `0, 0` (Roman, language-neutral) silently.

**Suggestion:** emit the Macintosh subtable only when a Mac CMap resource is actually supplied via `-cm` (or provide an option to omit it).

### 2. `-cl` is written into the language field without the spec-required +1

The spec (["Use of the language field in 'cmap' subtables"](https://learn.microsoft.com/en-us/typography/opentype/spec/cmap#use-of-the-language-field-in-cmap-subtables)) requires the Mac-platform language field to be the **Macintosh language ID plus one** (or 0 if language-neutral):

> "For 'cmap' subtables whose platform IDs are Macintosh, set this field to the Macintosh language ID of the 'cmap' subtable plus one, or to zero if the 'cmap' subtable is not language-specific. For example, a Mac OS Turkish 'cmap' subtable must set this field to 18, since the Macintosh language ID for Turkish is 17."

hotconv's internal convention follows this correctly — see the struct comment and the heuristics table:

- [map.cpp#L158-L159](https://github.com/adobe-type-tools/afdko/blob/ca7161c/c/addfeatures/hotconv/map.cpp#L158-L159): `/* Language field for Mac cmap (this is Mac language ID + 1, or 0 if unspecific language) */`
- [codepage.h#L20-L21](https://github.com/adobe-type-tools/afdko/blob/ca7161c/c/addfeatures/hotconv/codepage.h#L20-L21): the Turkish row stores `18` (= 17+1).

But the `-cl` option value bypasses that convention — it is passed through verbatim:

- [main.cpp#L273-L278](https://github.com/adobe-type-tools/afdko/blob/ca7161c/c/addfeatures/main.cpp#L273-L278): `convert.macLanguage = atoi(argv[++i]);`
- [cb.cpp#L767-L768](https://github.com/adobe-type-tools/afdko/blob/ca7161c/c/addfeatures/cb.cpp#L767-L768): `h->mac.cmapLanguage = macLanguage;`
- [cmap.cpp#L420](https://github.com/adobe-type-tools/afdko/blob/ca7161c/c/addfeatures/hotconv/cmap.cpp#L420): `fmt->language = h->language;`

…while both the usage text (*"-cl <integer> : Integer value for Mac cmap language ID. -1 means undefined."*) and the MakeOTF User Guide (*"specify the LanguageID for the Macintosh 'cmap' subtable"*) read as the plain Macintosh language ID. So `-cl 33` (Simplified Chinese) produces `language = 33`, which per spec decodes as language ID 32; the conforming value would be 34.

Note that the no-`-cl` default is unaffected: `convert` is zero-initialized, so builds without `-cl` get `language = 0` (language-neutral), which is spec-correct. Only explicitly passed `-cl` values are off by one.

**Suggestion:** if fixing in code, apply the `+1` only at the point where the `-cl` value is consumed ([main.cpp#L277](https://github.com/adobe-type-tools/afdko/blob/ca7161c/c/addfeatures/main.cpp#L273-L278), for values ≥ 0, preserving the documented `-1 = undefined` sentinel) — **not** in the shared write path: the same field is also fed by three already-spec-correct sources that must not be shifted (the zero default, the CID fallback `MAC_LANG_UNSPEC = 0`, and the heuristics table whose entries are already stored as ID+1). A blanket `+1` at write time would, among other things, turn every default build's `0` (neutral) into `1` (English-specific). Alternatively, fix by documentation only: state that `-cl` expects the raw field value (language ID + 1).

Minor related typo: the comment at [main.cpp#L273](https://github.com/adobe-type-tools/afdko/blob/ca7161c/c/addfeatures/main.cpp#L273) says `/* [-cl] Mac Adobe CMap script id */` — should read "language id".

skef:
## Summary

Two fixes for the Macintosh cmap subtable in CID-keyed font builds:

- **Suppress empty Mac cmap stub** when no Mac CMap resource is provided via `-cm`. Previously `makeCIDMaccmap()` was called unconditionally, emitting a useless platform-1 subtable mapping only .notdef. Now the subtable is only created when `-cm` data is available.
- **Fix `-cl` off-by-one**: apply the spec-required +1 to the user's value at parse time. The internal API convention (documented in `hotconv.h:380-381`) is "language ID + 1, or 0 if unspecific." All other paths honor this, but `-cl` passed through raw. So `-cl 33` (Simplified Chinese) produced `language=33` (decodes as ID 32 per spec); correct value is 34.
- Fix comment typo: `-cl` case said "script id", should be "language id".

Fixes #1842

## Questions for reviewer

1. **Backwards compatibility of suppression**: Could any existing CID workflow depend on the empty platform-1 stub being present? The spec discourages Mac cmap subtables and the stub maps nothing useful, but it's a behavior change. The warning message is retained (with updated text) so users will know what changed.

2. **Should the warning be downgraded to INFO?** The reporter notes it's confusing to warn about a missing resource for a subtable the user never asked for. Now that we suppress the subtable, the warning is purely informational ("we didn't make a Mac cmap because you didn't give us Mac CMap data").

3. **Should `-cl` document the change?** The fix means existing scripts passing `-cl` values will now get different (correct) output. Anyone who was compensating by passing the already-incremented value will now be off by one in the other direction. Probably fine since the old behavior was simply wrong, but worth noting.

## Test plan

- [x] Verified the change compiles
- [ ] Existing test suite
- [ ] Manual test with a CID font build with/without `-cm`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


skef:
@frankrolf what do you think of this?

frankrolf:
Perhaps @hatchzo is more qualified to make an assessment here?

hatchzo:
Regarding `-cl`, I do not think this option is used very often in Adobe’s CJK fonts or in Japanese CID-keyed fonts.

For Adobe-Japan1 ROS-based fonts, an actual Macintosh cmap subtable is included. However, for fonts with the Adobe-Identity-0 ROS, the `-cs` option creates a stub subtable. This has continued from Kazuraki, Adobe’s first Adobe-Identity-0 font, through today’s Source Han Sans and Momochidori. At the time, Ken Lunde wrote in a blog post that although he recognized that the Macintosh cmap subtable was no longer needed on OS X, he included a stub table for heuristic reasons.

https://ccjktype.fonts.adobe.com/2013/06/ai0-and-heuristics.html

I assume this may have been done with legacy compatibility in mind, such as environments that check for the presence of Macintosh cmap subtable, but I am not sure what the specific impact is.

